PR #1, opened 2019-10-01
Added --target option for overriding target triple

geky:
Helps with cross-compilation when the target triple does not match the
machine you're building on.

I'm new to rust so let me know if I'm doing anything non-idiomatic.


geky:
Ok, I also tried to add target auto-detection through `llvm-config --host-target`. No idea if this is idomatic Rust or not, feel free to criticize.

Others:
My understanding is that leaving the target blank makes LLVM pick the default target automatically. Is that incorrect?

(Will look at the other commits in detail later)

geky:
Ah, so it does! Dropped that logic for just omitting the field if target is not provided.

PR #2, opened 2019-10-01
Stubbed out MemorySize and MemoryGrow instructions

geky:
This is required to be able to compile WASI-SDK. I'm not sure what the correct behavior should be. Just having these as noops seems to work but I'm not sure that's expected. Maybe they should emit a function call to the expand_memory function?

Also I added tracking for the memory and table import sections, those are required if compiling incrementally.

### DIFF
--- a/src/codegen/block.rs
+++ b/src/codegen/block.rs
@@ -1020,6 +1020,13 @@ pub fn compile_block<'a, 'b>(
                 let v = stack.pop().unwrap();
                 store_val::<f64>(m_ctx, b, &mut stack, offset, v);
             }
+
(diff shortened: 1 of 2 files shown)


Others:
It's a little scary to me this works. Would it be difficult to just hook this up to the runtime? (Or at the very least, pop arguments and push dummy values.) I'm concerned this could corrupt the stack, since the semantics of WASM rely on these consuming/returning values.


https://github.com/WebAssembly/design/blob/master/Semantics.md

geky:
Closing in favor of https://github.com/gwsystems/aWsm/pull/14

PR #3, opened 2019-10-01
Added lazy memory backend for conservative memory allocation

geky:
Here is an attempt at the lazy memory backend. It implements Silverfish's memory backend by lazily allocating blocks of memory in a size of WASM_BLOCK_SIZE. These blocks are not allocated until the first write and read returns 0 on unallocated blocks. This lets us get below the 64KiB page limit built into WebAssembly.

The main downside is the overhead of byte-level access for everything. This is unfortunately required as WebAssembly does not guarantee integer alignment, which can cause accesses to cross block boundaries. There may be a more clever way to handle this.

Note! Another side-effect is that this memory backend can't provied the `get_memory_ptr_for_runtime` function as it takes memory handling past the scope of simple bounds checks. I have a fork of the runtime that does not need `get_memory_ptr_for_runtime`, but it is very hack so I will push it up as another PR.

### DIFF
--- a/runtime/memory/lazy.c
+++ b/runtime/memory/lazy.c
@@ -1,0 +1,143 @@
+#include "../runtime.h"
+
+#ifndef WASM_BLOCK_SIZE
+#define WASM_BLOCK_SIZE 4096
+#endif
+
+void **memory_blocks;
+u32 memory_size;
+
+void alloc_linear_memory() {
+    memory_blocks = malloc((starting_pages * WASM_PAGE_SIZE / WASM_BLOCK_SIZE)
+            * sizeof(void*));
+    memset(memory_blocks, 0, (starting_pages * WASM_PAGE_SIZE / WASM_BLOCK_SIZE)
+            * sizeof(void*));
+    memory_size = starting_pages * WASM_PAGE_SIZE;
+    printf("mem size: %d\n", memory_size);
+}
+
+void expand_memory() {
+    printf("mem size: %d -> %d\n", memory_size, memory_size + WASM_PAGE_SIZE);
+    // max_pages = 0 => no limit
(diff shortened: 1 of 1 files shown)


runtime/memory/lazy.c
```diff
@@ -0,0 +1,143 @@
...
+
+// FIXME: This can't be provided by this memory backend
+//INLINE char* get_memory_ptr_for_runtime(u32 offset, u32 bounds_check) {
+//    assert(memory_size > bounds_check && offset <= memory_size - bounds_check);
+//
+//    char* mem_as_chars = (char *) memory;
+//    char* address = &mem_as_chars[offset];
+//
+//    return address;
+//}
```
Others: Is this function not used by the runtime? / How does the runtime compile without this function?

runtime/memory/lazy.c
```diff
@@ -0,0 +1,143 @@
...
+INLINE i64 get_i64(i32 offset) {
+    return ((u64)(u8)get_i8(offset+0) << 0) |
+           ((u64)(u8)get_i8(offset+1) << 8) |
+           ((u64)(u8)get_i8(offset+2) << 16) |
+           ((u64)(u8)get_i8(offset+3) << 24) |
+           ((u64)(u8)get_i8(offset+4) << 32) |
+           ((u64)(u8)get_i8(offset+5) << 40) |
+           ((u64)(u8)get_i8(offset+6) << 48) |
+           ((u64)(u8)get_i8(offset+7) << 56);
+}
```
Others: Can you lift the bounds checks out here? No need to check the assert so many times (same for all the other repeated loads)

gparmer:
I believe this was addressed by @Others patches building up to the paper. Should we close this commit, or is it still relevant?

PR #6, opened 2020-01-06
WIP: Improve install script

bushidocodes:
My previous install script was terribly broken. This is slightly less so. WIP because I can't afford to spend more time on this while Awsm deadline is incoming. Pushed in case someone wants to help.


gparmer:
@Others This seems like a good addition to me. Is this antiquated given changes since the PR, or is it worth accepting?

Others:
@bushidocodes @gparmer my disposition is to merge this as-is. Seems like it could be useful, even if it may still be WIP. Good with you two?

PR #7, opened 2020-07-22
Awsm

gparmer:
Merge the serverless changes into mainline.


runtime/memory/64bit_nix.c
```diff
@@ -126,7 +126,9 @@ INLINE char* get_function_from_table(u32 idx, u32 type_id) {
 
     struct indirect_table_entry f = indirect_table[idx];
 
-    assert(f.type_id == type_id && f.func_pointer);
+    // FIXME: temporary as there is some issue in type id of a typecasted function pointer.
+//    assert(f.type_id == type_id);
+	   assert(f.func_pointer);
```
phanikishoreg, line 131: I should revert this and perhaps add a comment about a possible failure here in legacy C applications! This was only required with the GOCR test which was typecasting the function pointers to a different type and thus failing the type_id check!

phanikishoreg:
@gparmer Thank you so much for initiating this PR!

runtime/memory/64bit_nix.c
Others: Yeah def needs reverted. Commenting out this type check destroys sandbox security

phanikishoreg: Agreed!

Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Gregor Peach <gregorpeach@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-llvm-alt = { git = "https://github.com/Others/llvm-rs.git"}
+llvm-alt = { git = "https://github.com/gwsystems/llvm-rs.git"}
```
Others, line 8: What changes are we picking up here?

runtime/libc/libc_backing.c
```diff
@@ -516,7 +516,7 @@ u32 wasm_geteuid() {
 
 #define SYS_GET_TIME 228
 struct wasm_time_spec {
-    u32 sec;
+    u64 sec;
     u32 nanosec;
 };
```
Others, line 521: Can you cite where in musl this is set as 64 bit?

runtime/runtime.c
```diff
@@ -157,6 +157,10 @@ INLINE float f32_max(float a, float b) {
     return a > b ? a : b;
 }
 
+INLINE float f32_floor(float a) {
+    return floor(a);
```
Others, line 161: is this doing an implicit conversion to a double?

src/codegen/runtime_stubs.rs
```diff
@@ -72,9 +72,11 @@ pub const I64_REM: &str = "i64_rem";
 
 pub const F32_MIN: &str = "f32_min";
 pub const F32_MAX: &str = "f32_max";
+pub const F32_FLOOR: &str = "f32_floor";
 
 pub const F64_MIN: &str = "f64_min";
 pub const F64_MAX: &str = "f64_max";
+pub const F64_FLOOR: &str = "f64_floor";
```
Others, line 79: I'm a huge nitpicker, but can you alphabetize these blocks?

Cargo.toml
phanikishoreg: 1. I forked most of these repos in to gwsystems so other members in the group can work on this, like updating with upstream, other fixes, etc!
2. See [this](https://github.com/gwsystems/llvm-rs/commit/906d4830515efa1330fc7167d221278cb3cd1b0f). There is a small modification to use a portable datatype `libc::c_char` vs `i8` which seem to be defined differently on aarch64.

runtime/libc/libc_backing.c
phanikishoreg: In musl: https://git.musl-libc.org/cgit/musl/tree/arch/i386/bits/alltypes.h.in?id=38143339646a4ccce8afe298c34467767c899f51#n33

runtime/runtime.c
phanikishoreg: I think so, yeah! We could definitely use `floorf()` instead.

src/codegen/runtime_stubs.rs
phanikishoreg: DONE!

runtime/libc/libc_backing.c
Others: We use x86_64 as our base though. (This was obviously dumb, but it is what it is.) I guess it might be 64bits anyway? https://github.com/Others/musl/blob/master/arch/x86_64/bits/alltypes.h.in

phanikishoreg: u32 is likely `unsigned int` or `unsigned long`, and u64 is `unsigned long long` if I remember correctly. And the alltypes.h in x86_64 that you linked to also defines `time_t` as `long long`. 
Also to be clear, clock_gettime returns gibberish without this fix!

phanikishoreg:
@Others I guess I missed to answer your questions about `wasmception` and `llvm-alt`.
* `wasmception` changes are really to use the manually forked (forked from svn) repositories in gwsystems. I think it is only changing the Makefile!
* llvm-alt, I don't think I changed anything, again except that I made it use the gwsystems repository.

gparmer:
Seems like some changes are requested. Is this correct:

- floor -> floorf
- revert the removal of the typecheck on function pointer invocation.

Are those the only requested changes?

phanikishoreg:
I did revert the type check. I wasn't sure if `floorf` was actionable. 

We also have some merge conflicts apart from that, that Gregor mentioned on slack he'd resolve.

Others:
@phanikishoreg working on merge now, will fix floorf. did you run two threads in the same sandbox at the same time?

phanikishoreg:
Nope, all single threaded applications. I'm pretty sure we make that assumption in our musl port.

Reference: https://github.com/gwsystems/wasmception-musl/commit/d92987dfced63df9a0e8268b81ab7d9dee4312ae#diff-be6fae696397cfe7a11e9f75f9feaeaeR1

runtime/libc/libc_backing.c
phanikishoreg: Just noticed this: https://github.com/gwsystems/wasmception-musl/commit/d92987dfced63df9a0e8268b81ab7d9dee4312ae#diff-1f02b3cd6244cd1511b2020556d1c584R22

This diff suggests time_t was originally long in x86_64 but you seemed to have changed it to long long. That explains the bug, because that change wasn't propagated to the libc in awsm. I think you know best to fix this the right way, just sharing for more information!

Others:
Merge on the way. I'm not going to pull the submodule update yet: I'd like to try and remove our dependency on llvm-alt entirely. Probably a weekend of work.

Others:
Okay almost there :)

@phanikishoreg can you review my modifications?

@gparmer I noticed in the README, we instruct people to install then test with no disclaimers. I worry this is misleading, as the test script uses the development version (from the `target` directory). Would it be more clear to have separate "here is how you test" and "here is how you install" sections? Wondering if the script should grow an option to turn off the global install as well.

Cargo.lock
```diff
@@ -2,446 +2,456 @@
...
-version = "0.2.51"
+version = "0.2.73"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bd7d4bd64732af4bf3a67f367c27df8520ad7e230c5817b8ff485864d80242b9"
 
 [[package]]
 name = "llvm-alt"
 version = "0.5.0"
-source = "git+https://github.com/Others/llvm-rs.git#754fdf857f83e64d371a046dcb8fa3587a176cf8"
+source = "git+https://github.com/Others/llvm-rs.git#4f825ff10f7bdc20a3beb2bca648f72f0a0b1af7"
```
phanikishoreg, line 155: As you're not going for that gwsystems/llvm-rs, I think this change should be cherry-picked: https://github.com/gwsystems/llvm-rs/commit/906d4830515efa1330fc7167d221278cb3cd1b0f, so we don't break aarch64 build?

Others:
@phanikishoreg Okay I tried just swapping over to gwsystems/awsm, and it causes a ton of build errors on my end. It might be an easy fix, but it looks like it needs a chain of submodule updates. Would you mind making a follow up PR with just the submodule update so we can test that independently? We can merge this now and I can start working on ripping out llvm-alt concurrently with that PR?

phanikishoreg:
Done! I PRed to your fork of that one change, if you accept it, do you mind updating the cargo.lock in this? Thanks!

gparmer:
> @gparmer I noticed in the README, we instruct people to install then test with no disclaimers. I worry this is misleading, as the test script uses the development version (from the `target` directory). Would it be more clear to have separate "here is how you test" and "here is how you install" sections? Wondering if the script should grow an option to turn off the global install as well.

Great point. A different take in #13

Issue #9, opened 2020-07-23
LLVM ERROR: inconsistency in registered CommandLine options     


geky:
At the moment the dual dependencies on llvm-sys causes a strange error with LLVM:
``` bash
$ ./target/debug/silverfish               
: CommandLine Error: Option 'help-list' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options       
```

Some googling around shows this is an issue that can happen if libLLVM gets linked twice:
https://stackoverflow.com/questions/54273632/llvm-linking-commandline-error-option-help-list-registered-more-than-once-l

The dual dependencies against llvm-sys might be the issue? I'm not sure if there is a way to tell cargo to deduplicate the linking:
```
$ cargo tree
llvm-alt v0.5.0 (git+https://github.com/Others/llvm-rs.git#754fdf857f83e64d371a046dcb8fa3587a176cf8)     
├── cbox v0.3.0                                                                                          
│   └── libc v0.2.51 (*)                                                                                 
├── libc v0.2.51 (*)                                                                                     
└── llvm-sys v0.3.0 (git+https://github.com/Others/llvm-sys.rs#f4d6b7523e9c3e71a45aa8155f0ef0911689033a) 

llvm-sys v90.0.0    
└── libc v0.2.51 (*)
```

Others:
Probably suggests that removing llvm-alt is the path forward for #11

Others:
Potentially fixed when #18 is merged?

PR #10, opened 2020-07-23
Moved Cortex-M specific LLVM attrs behind features=cortex-m-attrs

geky:
This moves the Cortex-M-specific attributes behind the feature flag "cortex-m-attrs". Happen to change it if you prefer a different name.

Note the appropriate checks are already in place so that the attributes are only added if the target is "thumbv7em-none-unknown-eabi", but there is a separate issue with the dual dependency on llvm-sys that occurs at link-time. I've moved the dual-dependency behind the feature flag to avoid this.

Captured the info I know here: https://github.com/gwsystems/aWsm/issues/9

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,11 +6,14 @@ edition = "2018"
 
 [dependencies]
 flexi_logger = "0.14.4"
-llvm-sys = "90.0.0"
-llvm-alt = { git = "https://github.com/Others/llvm-rs.git"}
+llvm-sys = { version = "90.0.0", optional = true }
+llvm-alt = { git = "https://github.com/Others/llvm-rs.git" }
 log = "0.4.8"
 structopt = "0.3.2"
 wasmparser = "0.39.2"
+
+[features]
+cortex-m-attrs = ['llvm-sys']
 
 [profile.release]
 debug = true
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -1,8 +1,12 @@
 use std::cell::Cell;
+#[cfg(feature = "cortex-m-attrs")]
 use std::ffi::CString;
+#[cfg(feature = "cortex-m-attrs")]
 use std::mem;
 
-use llvm::{BasicBlock, Sub};
+use llvm::BasicBlock;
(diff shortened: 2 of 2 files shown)


geky:
Oh, that's interesting how your comment got moved here from the commit comment.

gparmer:
@Others @phanikishoreg You guys have more context about if this is the right path?

phanikishoreg:
Sorry, I don't! 

But I have many of those repos used in awsm forked in to gwsystems organization. 
https://github.com/gwsystems/llvm-rs

geky:
Just to clarify, it would still be an issue with both repos on gwsystems. The issue is the diamond dependency to libLLVM, which it seems like the libLLVM/Rust FFI layer can't handle? I'm sort of guessing.

```
aWsm ---------------> llvm-sys -> libLLVM
 '-----> llvm-rs  --> llvm-sys ----^
```

geky:
Oh, this may also be fixed by bring both llvm-sys dependencies in sync, maybe? Their versions are a tiny bit different (v90.0.0 vs v0.3.0). I tried just setting them to the same version but quite a bit of the API has changed.

Others:
I want to fix the underlying issue here I think. Let me explore that before I merge this

Others:
Okay my migration attempt is hitting some roadblocks, it's very annoying to switch over, and I need to play with it more. However, I think I have a better hack than hiding behind a feature.

`llvm-alt`, which we import as `llvm` has a module `ffi` wrapping `llvm_sys`
```
pub extern crate llvm_sys as ffi;
```
We can use that to replace our dependency on `llvm_sys`

However, some functions may be missing since that version is old. To fix this, some manual `extern` blocks may need to be added.

What do you think of that as a solution as an alternative to this PR? It would make the problem invisible to the end user, at the cost of some tech debt. (Debt that I'm working on removing...)

Others:
@geky forgot to tag you in the above

geky:
Sounds reasonable to me. I'm all in favor of a temporary solution.

And it sounds like this would get cleaned up in https://github.com/gwsystems/aWsm/issues/11 anyways?

geky:
Closing, this was fixed 👍

PR #12, opened 2020-07-26
Fix testing on Ubuntu 19.10

geky:
Here were the changes needed to get the code_benches running on Ubuntu 19.10, Rust 1.45.0, LLVM 9.0.0

Changes:
- Replaced shopt with python's glob.glob.

  Needed as Linux defaults to sh with shell=True. There is an executable argument to allow passing bash, but I'm not sure it's portable to other OSs

- Added -DSQLITE_MUTEX_NOOP and -ldl to get custom_sqlite compiling
- Disabled app_nn for now, depends on arm_math and won't work portably
- Added -I. to tinycrypt since it seems to use `<anglebrackets>` for its includes.

I also added a stub to tests/code_benches.rs (new file) so that `cargo test` runs code_benches/run.py. This is not needed so I can remove if you'd like.

~Also note this is currently built on https://github.com/gwsystems/aWsm/pull/10, since I needed it to get aWsm running. Happy to rebase once https://github.com/gwsystems/aWsm/issues/9 is resolved.~ Rebasing onto current fix, see thread

### DIFF
--- a/code_benches/run.py
+++ b/code_benches/run.py
@@ -113,10 +113,10 @@ programs = [
     Program("custom_sqlite", [], 2 ** 15, custom_arguments=["-DSQLITE_MUTEX_NOOP", "-ldl"]),
 
     # == Apps ==
-    # Program("app_nn", [], 2 ** 14, custom_arguments=["-std=c99", "-Wno-unknown-attributes", "-DARM_MATH_CM3", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/DSP/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/Core/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/NN/Include"]),
-#    Program("app_pid", ["-std=c++11", "-Wall"], 2 ** 8, custom_arguments=[], is_cpp=True),
-#    Program("app_tiny_ekf", ["-std=c++11", "-Wall"], 2 ** 14, custom_arguments=[], is_cpp=True),
-#    Program("app_tinycrypt", [], 2 ** 15 + 2**14, custom_arguments=[ "-Wall", "-Wpedantic", "-Wno-gnu-zero-variadic-macro-arguments", "-std=c11", "-I/Users/peachg/Projects/silverfish/code_benches/app_tinycrypt/", "-DENABLE_TESTS", "-I."]),
+    #Program("app_nn", [], 2 ** 14, custom_arguments=["-std=c99", "-Wno-unknown-attributes", "-DARM_MATH_CM3", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/DSP/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/Core/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/NN/Include"]),
+    Program("app_pid", ["-std=c++11", "-Wall"], 2 ** 8, custom_arguments=[], is_cpp=True),
+    Program("app_tiny_ekf", ["-std=c++11", "-Wall"], 2 ** 14, custom_arguments=[], is_cpp=True),
+    Program("app_tinycrypt", [], 2 ** 15 + 2**14, custom_arguments=[ "-Wall", "-Wpedantic", "-Wno-gnu-zero-variadic-macro-arguments", "-std=c11", "-DENABLE_TESTS", "-I."]),
     # Program("app_v9", [], 2 ** 18, custom_arguments=[], do_lto=False),
 
     # == MiBench ==
--- a/code_benches/run.py
+++ b/code_benches/run.py
@@ -177,11 +177,14 @@ def compile_to_executable(program):
     if ENABLE_DEBUG_SYMBOLS:
         opt += " -g"
     if program.is_cpp:
-        sp.check_call("clang++ {} -lm {} {} -o bin/{}".format(program.custom_arguments, opt, program.sources(), program.name), shell=True, cwd=program.name)
+        clang = "clang++"
     else:
-        sp.check_call("clang {} -lm {} {} -o bin/{}".format(program.custom_arguments, opt, program.sources(), program.name), shell=True, cwd=program.name)
-        # sp.check_call("clang {} -lm {} *.c -o bin/{}".format(program.custom_arguments, opt, program.name), shell=True, cwd=program.name)
+        clang = "clang"
 
+    command = "{clang} {args} -lm {opt} {sources} -o bin/{pname}" \
+        .format(clang=clang, args=program.custom_arguments, opt=opt, sources=program.sources(), pname=program.name)
+    print(command)
+    sp.check_call(command, shell=True, cwd=program.name)
 
 # Compile the C code in `program`'s directory into WASM
 def compile_to_wasm(program):
--- a/code_benches/run.py
+++ b/code_benches/run.py
@@ -3,6 +3,8 @@ import os
 import subprocess as sp
 import sys
 import timeit
+import glob
+import itertools as it
 
 # Note: This is a major configuration option, you probably want to set this if you're doing anything non-trivial
 SILVERFISH_TARGET = None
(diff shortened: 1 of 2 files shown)


gparmer:
This seems like only value added to me. @Others and @phanikishoreg will likely have an opinion.

geky:
EDIT:
> I also added a stub to tests/code_benches.rs (new file) so that cargo test runs `code_benches/run.py`. This is **not** needed so I can remove if you'd like.

I missed the most important word in that sentence...

code_benches/run.py
```diff
@@ -76,13 +96,13 @@ def __str__(self):
                               "-Wno-shift-negative-value"]),
     Program("custom_matrix_multiply", [], 2 ** 14),
     Program("custom_memcmp", [], 2 ** 14),
-    Program("custom_sqlite", [], 2 ** 15),
+    Program("custom_sqlite", [], 2 ** 15, custom_arguments=["-pthreads", "-ldl"]),
```
phanikishoreg, line 99: Not sure if `-pthreads` should be used though, as we don't support multi-threading within a Wasm application.

phanikishoreg:
Sorry again, I don't have much context into a lot of these changes. But I can only say "fix testing on Ubuntu 19.10" should be useful changes to incorporate. However, I'm surprised that so many changes are required to be able to test on 19.10, especially if we had it running on 18.04 just fine without these.

code_benches/run.py
geky: Ah, it looks like sqlite was just using -pthreads for the mutex definitions. Alternatively it looks like I can use -DSQLITE_MUTEX_NOOP, will change.

Others:
I got all the I newly commented ones to work on x86 I think? Wondering what's going on Ubuntu

geky:
Hmm, I'm really not sure. Most of these looks like they could be compiler version differences. I'm using clang v9.0.0, but maybe the 18.04 machine had a different native-clang installed?

geky:
Changes:
- Adopted -DSQLITE_MUTEX_NOOP instead of -pthreads
- Rebased onto https://github.com/gwsystems/aWsm/pull/18

Others:
I merged #18. @geky would you mind merging from master here? Github isn't rendering your changes separate from mine :(

geky:
Oh yeah, I don't know what I was expecting. Should be updated now 👍

code_benches/run.py
```diff
@@ -13,13 +17,19 @@
...
 BENCH_ROOT = os.getcwd()
 # Absolute path to the `silverfish` directory
 ROOT_PATH = os.path.dirname(BENCH_ROOT)
 
 RUNTIME_PATH = ROOT_PATH + "/runtime"
-SILVERFISH_PATH = ROOT_PATH + "/target/release/silverfish"
+if "--release" in sys.argv:
+    SILVERFISH_PATH = ROOT_PATH + "/target/release/silverfish"
+else:
+    SILVERFISH_PATH = ROOT_PATH + "/target/debug/silverfish"
```
Others, line 32: I'm not sure if this is the right default. Maybe this is just my workflow, but I usually build silverfish in release mode, and this means more typing for me :P

Maybe this is too magic, but maybe we could use the newest of `target/release/silverfish` and  `target/debug/silverfish`?  (using whatever one is existant  if only one exists) Then add manual `--release` and `--debug` flags to override this magic choice? Let me know what you think on this--I'm happy to write that code as a follow up PR myself

Ignoring that, can you add validation that there are no useless args in argv? Don't want someone typoing `relese` or something like that

code_benches/run.py
```diff
@@ -61,6 +71,16 @@ def __init__(self, name, parameters, stack_size, custom_arguments=None, do_lto=T
...
 
+    def sources(self):
+        # glob here, avoids issues with glob expansion in shell
+        if self.is_cpp:
+            patterns = ["*.c", "*.cpp"]
+        else:
+            patterns = ["*.c"]
+
+        sources = it.chain.from_iterable(glob.glob(os.path.join(self.name, pattern)) for pattern in patterns)
+        return " ".join(source[len(self.name)+1:] for source in sources)
```
Others, line 82: just import `chain` directly

also, can you split this into more variables, had some trouble parsing this at first glance

code_benches/run.py
```diff
@@ -76,13 +96,13 @@ def __str__(self):
                               "-Wno-shift-negative-value"]),
     Program("custom_matrix_multiply", [], 2 ** 14),
     Program("custom_memcmp", [], 2 ** 14),
-    Program("custom_sqlite", [], 2 ** 15),
+    Program("custom_sqlite", [], 2 ** 15, custom_arguments=["-DSQLITE_MUTEX_NOOP", "-ldl"]),
```
Others, line 99: Note to self: need to test on OSX

code_benches/run.py
```diff
@@ -76,13 +96,13 @@ def __str__(self):
...
-    Program("app_pid", ["-std=c++11", "-Wall"], 2 ** 8, custom_arguments=[], is_cpp=True),
-    Program("app_tiny_ekf", ["-std=c++11", "-Wall"], 2 ** 14, custom_arguments=[], is_cpp=True),
-    Program("app_tinycrypt", [], 2 ** 15 + 2**14, custom_arguments=[ "-Wall", "-Wpedantic", "-Wno-gnu-zero-variadic-macro-arguments", "-std=c11", "-I/Users/peachg/Projects/silverfish/code_benches/app_tinycrypt/", "-DENABLE_TESTS"]),
+    # Program("app_nn", [], 2 ** 14, custom_arguments=["-std=c99", "-Wno-unknown-attributes", "-DARM_MATH_CM3", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/DSP/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/Core/Include", "-I/Users/peachg/Projects/CMSIS_5_NN/CMSIS_5/CMSIS/NN/Include"]),
+#    Program("app_pid", ["-std=c++11", "-Wall"], 2 ** 8, custom_arguments=[], is_cpp=True),
+#    Program("app_tiny_ekf", ["-std=c++11", "-Wall"], 2 ** 14, custom_arguments=[], is_cpp=True),
+#    Program("app_tinycrypt", [], 2 ** 15 + 2**14, custom_arguments=[ "-Wall", "-Wpedantic", "-Wno-gnu-zero-variadic-macro-arguments", "-std=c11", "-I/Users/peachg/Projects/silverfish/code_benches/app_tinycrypt/", "-DENABLE_TESTS", "-I."]),
```
Others, line 105: I think the problem here is the hardcoded include paths. Could you see if fixing those means we can uncomment these tests?

code_benches/run.py
geky: That is something I can't really do so I'll leave it up to you :)

code_benches/run.py
geky: Oh, that may be the problem. I will try.

From some of our previous discussions it sounded like the chance of arm_math.h working on x86_64 is unlikely, or do you expect CMSIS_NN will work locally?

code_benches/run.py
geky: Ah, I should have noted the default change.

I you run `cargo test` this will do the "right thing" based on if you passed --release or not.

I should be able to add a timestamp-based check (double-checking is that what you are asking for?), it will just take more perusing of Rust's APIs. The override flags make sense and should be easy to add in both the script and `cargo test`.

---

I can add the argv check. My default is to be the least invasive, but making it user-friendly is a good idea.

code_benches/run.py
geky: Can import chain directly 👍 

Do you have a suggestion where to split this?

I can store a temporary like:
```
paths = (os.path.join(self.name, pattern) for pattern in pattern)
sources = it.chain.from_iterable(glob.glob(path) for path in paths))
```

Does that accomplish what you're looking for?

Unfortunately handling the results from the standard python APIs is rather annoying.

code_benches/run.py
Others: I think it mostly works. Works on my machine at least :P

geky:
Sorry about the delay on this, just pushed up changes per feedback.

Added the "use most recent silverfish" behavior, with override flag and cargo test overriding based on its configuration.

I was able to get app_pid, app_tiny_ekf, and app_tinycrypt back to working by fixing the include paths as you mentioned, but not app_nn as it depends on CMSIS_NN.

geky:
Sure thing 👍 Feel free to consume how you'd like

PR #14, opened 2020-07-28
 Added codegen for memory.size and memory.grow instructions 

geky:
Creating PR for review, currently building off of https://github.com/gwsystems/aWsm/pull/12, happy to rebase on what changes goes in. Only the last commit should be relevant: https://github.com/gwsystems/aWsm/commit/fc2bbe57ffc9304b28dab29b3b3dcd3dc5e25be0

I tried to follow the implementation of instruction->runtime function instructions, so hopefully this is better than my previous attempt (https://github.com/gwsystems/aWsm/pull/2).

Currently linking to `memory_grow` and `memory_size_`. Need to fix the name conflict with the `memory_size` variable, though happy to take alternative names.

Have we considered prefixing all of the runtime functions with something like `awsmr_memory_grow`?

### DIFF
--- a/runtime/memory/cortex_m_wrapping.c
+++ b/runtime/memory/cortex_m_wrapping.c
@@ -27,6 +27,19 @@ void expand_memory() {
     memory_size += WASM_PAGE_SIZE;
 }
 
+i32 instruction_memory_size() {
+    return memory_size / WASM_PAGE_SIZE;
+}
+
+i32 instruction_memory_grow(i32 count) {
+    i32 prev_size = instruction_memory_size();
+    for (int i = 0; i < count; i++) {
+        expand_memory();
+    }
+
+    return prev_size;
+}
+
 INLINE char* get_memory_ptr_for_runtime(u32 offset, u32 bounds_check) {
     silverfish_assert(offset <= memory_size - bounds_check);
 
--- a/runtime/memory/generic.c
+++ b/runtime/memory/generic.c
@@ -22,6 +22,19 @@ void expand_memory() {
     memory_size += WASM_PAGE_SIZE;
 }
 
+i32 instruction_memory_size() {
+    return memory_size / WASM_PAGE_SIZE;
+}
+
+i32 instruction_memory_grow(i32 count) {
+    i32 prev_size = instruction_memory_size();
+    for (int i = 0; i < count; i++) {
+        expand_memory();
+    }
+
+    return prev_size;
+}
+
 INLINE char* get_memory_ptr_for_runtime(u32 offset, u32 bounds_check) {
     silverfish_assert(memory_size > bounds_check && offset <= memory_size - bounds_check);
 
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -440,6 +440,9 @@ pub enum Instruction {
 
     F64Load { flags: u32, offset: u32 },
     F64Store { flags: u32, offset: u32 },
+
+    MemorySize,
+    MemoryGrow,
 }
 
 impl<'a> From<&'a Operator<'a>> for Instruction {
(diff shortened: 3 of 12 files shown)


Others:
I would prefix both these functions with "instruction_" for now. I see this builds off your other PR

Sorry for not being fast with the reviews on these, working very slowly on switching away from llvm_alt

geky:
Will update.

No worries on the speed. I'm used to PRs dripping through review processes like a coffee filter. Once I can get changes in PR form and presented to you guys I'm happy to have it sit until needed.

It's unfortunate GitHub doesn't render small inter-dependent PRs nicer.

Others:
Reading through these changes--all looks good. Will rereview after the PRs this one is based off are merged :)

Others:
@geky can you rebase this off master?

geky:
Updated! This has also been rolled up into https://github.com/gwsystems/aWsm/pull/19, which has a number of other small changes. Feel free to take this in as a part of https://github.com/gwsystems/aWsm/pull/19 if that is easier.

src/wasm.rs
```diff
@@ -971,7 +977,12 @@ impl WasmModule {
                             mutable: global_ty.mutable,
                         });
                     }
-                    e => panic!("Have not implemented import section entry type {:?}", e),
+                    ImportSectionEntryType::Memory(memory_ty) => {
+                        self.memories.push(*memory_ty);
+                    }
+                    ImportSectionEntryType::Table(table_ty) => {
+                        self.tables.push(*table_ty);
+                    }
```
Others, line 985: Kinda nervous about this, as we don't actually support additional memories or tables. I suppose I can't think of anywhere else in the code we assume this (or at least nowhere that'd cause subtle unsafety)

src/wasm.rs
```diff
@@ -753,6 +756,9 @@ impl<'a> From<&'a Operator<'a>> for Instruction {
                 offset: memarg.offset,
             },
 
+            Operator::MemorySize { .. } => Instruction::MemorySize,
+            Operator::MemoryGrow { .. } => Instruction::MemoryGrow,
```
Others, line 760: Are there no useful parameters to pipe down?

runtime/memory/generic.c
```diff
@@ -22,6 +22,19 @@ void expand_memory() {
...
+
+i32 instruction_memory_grow(i32 count) {
+    i32 prev_size = instruction_memory_size();
+    for (int i = 0; i < count; i++) {
+        expand_memory();
+    }
+
+    return prev_size;
+}
+
```
Others, line 37: In retrospect, I really should've abstracted this repeated code out... this new addition is in line with everything else tho, so keep it this way

runtime/memory/cortex_m_wrapping.c
```diff
@@ -27,6 +27,19 @@ void expand_memory() {
     memory_size += WASM_PAGE_SIZE;
 }
 
+i32 instruction_memory_size() {
+    return memory_size / WASM_PAGE_SIZE;
+}
```
Others, line 32: We need to be careful here. If you look at the mmap implementation, technically there is some support for MMAP granularity smaller than wasm page size granularity. This could lead to surprising behavior from using this (seemingly simple) function, and we should at least document that.

Long term we probably should decide if we want to support the mmap hack.

src/wasm.rs
geky: Nope, the Wasm spec has a byte reserved for "memory index" which is currently unused
https://webassembly.github.io/spec/core/binary/instructions.html#memory-instructions

> Note
>
> In future versions of WebAssembly, the additional zero bytes occurring in the encoding of the memory.size
> and memory.grow instructions may be used to index additional memories.

You could pipe down this byte, but then the unused variable would move into more places in the code base.

src/wasm.rs
geky: I believe these were needed because clang was emitting empty imports. They were empty, but aWsm was failing to parse the empty stubs.

Though I will admit there was a bit of poking-the-code-until-it-works here.

geky: It does match the `process_memory_section` function's behavior:
https://github.com/gwsystems/aWsm/blob/59e68b4dd1c1a9fb19df1c09f2301d0407302ac9/src/wasm.rs#L938

PR #18, opened 2020-07-29
Remove direct llvm-sys dependency, cleanup imports

Others:
Should fix any double-linking problems at the cost of a bit of tech debt. Alternative to #10.

I think this probably is better than hiding features behind a feature flag. @geky can you test whether this fixes your linking problems, the way #10 did?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,12 +57,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cc"
-version = "1.0.58"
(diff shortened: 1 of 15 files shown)


geky:
Can confirm this works for me 👍 

Even if it's a temporary fix, it's an improvement over both master and the feature-flag option, so I'm all in favor of bringing this in.

geky:
closes #10

PR #19, opened 2020-08-09
Add libc-backing support for WASI-SDK (v8)

geky:
The intent of this PR is to add initial support for WASI SDK. WASI SDK has a few benefits for Cortex-M development, such as allowing garbage collection of the libc-hooks. WASI SDK also provides prebuilt libc binaries, which is very nice for first time users. Also Wasmception has been deprecated in favor of WASI SDK, so would be a good idea to eventually migrate.

This PR has a few concessions:
1. Stuck on WASI SDK v8. Moving beyond v8, the Wasm backend for LLVM starts to emit `if` instructions, which aWsm currently doesn't support, and I'm not confident enough with the inner-workings of aWsm to implement branching instructions :)
2. Not an exhaustive implementation. Currently the WASI API is "best effort" similar to aWsm's level of instruction support. One part of this is that the WASI API is still a moving target, so it may be a waste of time to fully fill out the API (you may notice the API is prefixed with "wasi_unstable_"). The current API is plenty on Cortex-M, and a bit more than enough to get code_benches passing.

Changes:
- Added `memory.size` and `memory.grow` instructions. WASI SDK uses these instead of a mmap hook.
- Added `copysign/ceil/nearest` float instructions. The updated LLVM Wasm backend started emitting these. I tried to mirror the implementation of similar float instructions.
- Added wasi_sdk_backing.c and renamed wasmception_backing.c. These provide the backing for both libcs and can be chosen similar to the memory/*.c options.
- Added --wask-sdk and --wasmception options to code_benches/run.py. Also adopted argparse there as the argument handling seems to have gotten complex enough. By default run.py chooses the most recent one that exists.

Note for posterity:
- The WASI API is actually very frustrating to use with the expectations of stability. Marking the entire API as "unstable" is technically correct, but practically: unhelpful. However, the WASI-SDK (separate repo) does have easy to navigate releases.

  This leads to a bit of a strange situation where the best resource for the state of the WASI API at WASI-SDK v8 is this file at this hash:
  https://github.com/WebAssembly/wasi-libc/blob/ec3ee5e98541a89b9db016005fa0e9f0287fd755/libc-bottom-half/headers/public/wasi/core.h
  
Anyways, @Others, let me know how this looks


geky:
Sorry about the delay on this. This PR has a number of different changes in it, let me know if you want me to split it into multiple PRs.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -0,0 +1,339 @@
...
+    WASI_RIGHT_PATH_SYMLINK             = 0x0000000001000000ULL,
+    WASI_RIGHT_PATH_REMOVE_DIRECTORY    = 0x0000000002000000ULL,
+    WASI_RIGHT_PATH_UNLINK_FILE         = 0x0000000004000000ULL,
+    WASI_RIGHT_POLL_FD_READWRITE        = 0x0000000008000000ULL,
+    WASI_RIGHT_SOCK_SHUTDOWN            = 0x0000000010000000ULL,
+};
+
+
+// file operations
+i32 wasi_unstable_path_open(
```
gparmer, line 122: like the `_unstable_` namespacing.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -0,0 +1,339 @@
...
+
+i32 wasi_unstable_fd_read(i32 fd, i32 iov_offset, i32 iovcnt, i32 nread_off) {
+    i32 sum = 0;
+    struct wasi_iovec *iov = get_memory_ptr_void(iov_offset, iovcnt * sizeof(struct wasi_iovec));
+
+    for (int i = 0; i < iovcnt; i++) {
+        void* ptr = get_memory_ptr_void(iov[i].base_offset, iov[i].len);
+        ssize_t res = read(fd, ptr, iov[i].len);
+        if (res == -1) {
+            return -errno;
```
gparmer, line 250: I don't know if WASI errnos map 1:1 to POSIX. Hope so!

runtime/memory/64bit_nix.c
```diff
@@ -37,6 +37,19 @@ void expand_memory() {
     memory_size += WASM_PAGE_SIZE;
 }
 
+i32 instruction_memory_size() {
+    return memory_size / WASM_PAGE_SIZE;
```
gparmer, line 41: Assert that `memory_size` is a multiple of `WASM_PAGE_SIZE` so that we can make sure we don't have round down errors here?

gparmer: Offhand comment: Seems strange that this instruction returned a signed value.

runtime/memory/64bit_nix.c
```diff
@@ -37,6 +37,19 @@ void expand_memory() {
...
+    for (int i = 0; i < count; i++) {
+        expand_memory();
+    }
+
+    return prev_size;
+}
+
 INLINE char* get_memory_ptr_for_runtime(u32 offset, u32 bounds_check) {
     // Due to how we setup memory for x86, the virtual memory mechanism will catch the error, if bounds < WASM_PAGE_SIZE
     assert(bounds_check < WASM_PAGE_SIZE || (memory_size > bounds_check && offset <= memory_size - bounds_check));
```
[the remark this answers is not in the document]
gparmer: Put another way: we might want to (in the future) create an api for sandbox termination. This would enable environments with more than one sandbox in the process (e.g. sledge)

runtime/memory/cortex_m_wrapping.c
```diff
@@ -27,6 +27,19 @@ void expand_memory() {
     memory_size += WASM_PAGE_SIZE;
 }
 
+i32 instruction_memory_size() {
+    return memory_size / WASM_PAGE_SIZE;
+}
+
+i32 instruction_memory_grow(i32 count) {
```
gparmer, line 34: This feels pretty generic, but is replicated across backends. I wonder if there's an opportunity for more code sharing. Perhaps a future issue?

code_benches/run.py
```diff
@@ -8,14 +9,30 @@
...
+    help="Use Wasmception for the WebAssembly libc. Defaults to most recent that exists.")
+parser.add_argument("--wasi-sdk", action='store_true',
+    help="Use WASI-SDK for the WebAssembly libc. Defaults to most recent that exists.")
+parser.add_argument("-o", "--output", default="benchmarks.csv",
+    help="Destination csv file to write benchmark results. Defaults to %(default)r.")
+args = parser.parse_args()
+
 # Note: This is a major configuration option, you probably want to set this if you're doing anything non-trivial
-SILVERFISH_TARGET = None
+SILVERFISH_TARGET = args.target
```
Others, line 29: This allows an empty arg, right?

code_benches/run.py
```diff
@@ -27,32 +44,59 @@
...
+    *best, _ = max(paths, key=getmtime_or_zero)
+    return best
+
+SILVERFISH_RELEASE_PATH = ROOT_PATH + "/target/release/silverfish"
+SILVERFISH_DEBUG_PATH = ROOT_PATH + "/target/debug/silverfish"
+
+SILVERFISH_PATH, = bestpath([
+    (SILVERFISH_RELEASE_PATH, args.release),
+    (SILVERFISH_DEBUG_PATH,   args.debug),
+])
```
Others, line 73: What happens if someone passes in both `--release` and `--debug`?

code_benches/run.py
```diff
@@ -27,32 +44,59 @@
...
 WASMCEPTION_PATH = ROOT_PATH + "/wasmception"
+WASMCEPTION_CLANG = WASMCEPTION_PATH + "/dist/bin/clang"
+WASMCEPTION_SYSROOT = WASMCEPTION_PATH + "/sysroot"
+WASMCEPTION_FLAGS = "--target=wasm32-unknown-unknown-wasm -nostartfiles -O3 -flto"
+WASMCEPTION_BACKING = "wasmception_backing.c"
+
+WASI_SDK_PATH = ROOT_PATH + "/wasi-sdk"
+WASI_SDK_CLANG = WASI_SDK_PATH + "/bin/clang"
+WASI_SDK_SYSROOT = WASI_SDK_PATH + "/share/wasi-sysroot"
+WASI_SDK_FLAGS = "--target=wasm32-wasi -mcpu=mvp -nostartfiles -O3 -flto"
```
Others, line 84: are there other options for cpu?

code_benches/run.py
```diff
@@ -105,17 +149,18 @@ def sources(self):
...
-            custom_arguments=["-Wno-incompatible-library-redeclaration", "-Wno-implicit-function-declaration",
-                              "-Wno-shift-negative-value"]),
+    #Program("custom_libjpeg", [], 2 ** 15,
+    #        custom_arguments=["-Wno-incompatible-library-redeclaration", "-Wno-implicit-function-declaration",
+    #                          "-Wno-shift-negative-value"]),
     Program("custom_matrix_multiply", [], 2 ** 14),
     Program("custom_memcmp", [], 2 ** 14),
-    Program("custom_sqlite", [], 2 ** 15, custom_arguments=["-DSQLITE_MUTEX_NOOP", "-ldl"]),
+    # TODO need to remove dependency on posix headers
+    #Program("custom_sqlite", [], 2 ** 15, custom_arguments=["-DSQLITE_MUTEX_NOOP", "-ldl"]),
```
Others, line 158: Broken with wasi?

runtime/libc/env.c
```diff
@@ -147,3 +145,15 @@ env_getcycles(void)
...
+// TODO: Do a fair comparison between musl and wasm-musl
+INLINE double
+env_sin(double d) {
+    return sin(d);
+}
+
+INLINE double
+env_cos(double d) {
+    return cos(d);
+}
```
Others, line 159: This is needed for wasi??

runtime/libc/wasi_sdk_backing.c
```diff
@@ -0,0 +1,339 @@
...
+    set_i64(time_off, (uint64_t)tp.tv_sec*1000000000ULL + (uint64_t)tp.tv_nsec);
+    return 0;
+}
+
+// process operations
+__attribute__((noreturn))
+void wasi_unstable_proc_exit(i32 exitcode) {
+    exit(exitcode);
+}
+
```
Others, line 339: I can't proof these functions right now, too tired 😨  Will look in second review

Moving into apartment right now

runtime/memory/64bit_nix.c
Others: I don't think signed vs unsigned matters for correctness (WASM has no type level distinction), maybe u32 better for clearness?

runtime/memory/64bit_nix.c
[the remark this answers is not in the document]
Others: @gparmer I think we'd need to add special handling in `aWsm` for such a use case, so maybe we could do a different API? Creating branches manually in `aWsm` is tricky, so it might be better to redo a different memory API for that use case

runtime/memory/cortex_m.c
```diff
@@ -32,6 +32,19 @@ void expand_memory() {
...
+}
+
+i32 instruction_memory_grow(i32 count) {
+    i32 prev_size = instruction_memory_size();
+    for (int i = 0; i < count; i++) {
+        expand_memory();
+    }
+
+    return prev_size;
+}
```
Others, line 46: This seems very unlikely to work on cortex_m

runtime/memory/cortex_m_wrapping.c
Others: I think we can address in this PR. Maybe @geky can lift these to `runtime.c`?

runtime/runtime.h
```diff
@@ -99,6 +99,20 @@ void alloc_linear_memory();
...
+INLINE i64 get_i64(u32 offset);
+INLINE void set_i8(u32 offset, i8 v);
+INLINE void set_i16(u32 offset, i16 v);
+INLINE void set_i32(u32 offset, i32 v);
+INLINE void set_i64(u32 offset, i64 v);
+
+INLINE float get_f32(u32 offset);
+INLINE double get_f64(u32 offset);
+INLINE void set_f32(u32 offset, float);
+INLINE void set_f64(u32 offset, double);
```
Others, line 114: What does adding these actually get us? Typechecking?

src/wasm.rs
```diff
@@ -971,7 +989,12 @@ impl WasmModule {
                             mutable: global_ty.mutable,
                         });
                     }
-                    e => panic!("Have not implemented import section entry type {:?}", e),
+                    ImportSectionEntryType::Memory(memory_ty) => {
+                        self.memories.push(*memory_ty);
+                    }
+                    ImportSectionEntryType::Table(table_ty) => {
+                        self.tables.push(*table_ty);
+                    }
```
Others, line 997: Does `aWsm` import tables and memories?

runtime/libc/wasi_sdk_backing.c
geky: Wasn't me! This is the prefix WASI-SDK uses for its imports.

runtime/memory/64bit_nix.c
geky: Hmm, I think it's for consistency with memory.grow, which can return negative error codes. Since memory.grow uses i32, memory.size can't use the range >2^31 anyways.

Would we ever expect memory.size to return an error? The Wasm spec suggests no. I can change it next batch of changes if you all think we should.

Spec here:
https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-size

geky: Just noticed @gparmer's original comment. Can add assert 👍

code_benches/run.py
geky: Yep, if omitted argparse pases `None` so it should be the same behavior as before. I'm currently using the default (`None`). I just updated it since it seemed like a used config option.

code_benches/run.py
geky: Dunno. Probably breaks.

I didn't think it was worth enforcing, but I can look into how to enforce exclusivity if you think it is worth it.

geky: Do you know of an argparse option that would help?

code_benches/run.py
geky: Ah yes, this benchmark was using a number of Linux-specific headers (maybe based on OS checks), that wasi-sdk rejects while wasmception lets compile.

I haven't looked that much into it, but I think it's related to wasmception being based on syscalls whereas wasi-sdk is based on Wasm imports. wasmception's syscalls would turn into a runtime failure if called, whereas wasi-sdk's imports turn into a link-time failure in aWsm.

Do you all have experience configuring sqlite?

runtime/libc/env.c
geky: I'm not really sure. Where do the env_* functions get used?

These two were previously in libc_backing.c, instead of duplicating them in both wasmception_backing.c and wasi_sdk_backing.c, I moved them into env.c which looks to be where the other env_* functions are located.

runtime/runtime.h
geky: I used set/get_i32/64 as a part of the wasi-sdk implementation. Often the WASI API writes a 32-bit value to an address instead of returning it. So it was either declare these here, or declare them in wasi_sdk_backing.c.

I figured declaring these here would be better as it does the type checking (which is why a number of the memory files have been updated to use unsigned offsets).

runtime/libc/wasi_sdk_backing.c
geky: Hmm, initially I just carried over the errno handling from the wasmception_backing.c. But the implying the WASI API may mean these do need use a stricter conversion. I will update. It would get optimized out if they are identical anyways.

src/wasm.rs
geky: So! It does if you incrementally compile.

```
a.c --> a.wasm.o -.
b.c --> b.wasm.o -+-> combined.wasm
c.c --> c.wasm.o -'
            ^- These all used table/memory imports
```

It doesn't come up for the code_benches, but was a problem when I was applying aWsm to an existing project+build system.

code_benches/run.py
geky: Ah! At the time of writing this there are three:
- mvp (minimal viable product)
- bleeding-edge
- generic

Here:
https://github.com/llvm/llvm-project/blob/8c3e6af71b8e827655c83997747a2042feddc845/clang/lib/Basic/Targets/WebAssembly.cpp#L33-L34

I chose mvp as I know aWsm currently is inexaustive. I don't think there's actually a difference between mvp and generic at the moment, but there's a lot of proposed movement on Wasm that may cause issues.

A handful of non-trapping float instructions (fairly inconsequential) were added to the default cpu here:
https://reviews.llvm.org/D77908

There's also the recently added multi-valued return values that got added to Wasm recently. This may or may not be adopted in the default CPU. Related issue:
https://github.com/WebAssembly/wasi-sdk/issues/122

If someone wants to go through and make sure Wasm is covered exhaustively we could adopt -mcpu=default, but I worry this is a moving target.

runtime/libc/wasi_sdk_backing.c
geky: That's a good reason to take your time! 😄

runtime/memory/cortex_m.c
geky: The cortex_m implementation of expand_memory does move the memory limit it seems like. If it runs out of space in the configured buffer it will assert.

runtime/memory/cortex_m_wrapping.c
geky: Assuming expand_memory, memory_size, and WASM_PAGE_SIZE are all available globally I can move these to runtime.c.

Though I wonder if memory_grow/memory_size would be a better interface than expand_memory/memory_size. They are isomorphic to expand_memory/memory_size, but are closer the Wasm specification.

src/wasm.rs
geky: Second pass note: This seems to only happen if you incrementally compile _without_ LTO. With LTO the intermediaries are just LLVM IR blobs.

So it's uncommon, but I have locally tested this implementation so I figured it was worth upstreaming.

runtime/memory/64bit_nix.c
geky: Second pass: I think rounding down is actually the correct behavior. Nothing in the memory instructions strictly requires the page alignment. The only mechanisms available right now are limited to page boundaries, but if you provided something like a custom sbrk as a separate import (is this what wasmception does?), you could expand your memory by a non-page amount. This would work as long as the memory backend supports it.

If, say, you had 1.5 pages allocated, it seems accurate for `memory.size` to return 1. The code would then know you have at least 1 page. It might cause problems for users experimenting with non-page memory sizes if the `memory.size` instruction asserted.

geky:
Ok, sorry about being slow to update this! I've just pushed some changes based on your guys feedback:
- Moved memory instruction implementations into runtime.c
- Added proper conversion from system errno -> WASI errno
- Added asserts to check against incorrect argument combos (--debug+--release, --wasi-sdk+--wasmception)

Also added a small bit of test filtering so you can run a subset of the code_benches. This is useful for sanity checking minor changes without needing to run the whole bench-suite:
``` bash
./code_benches/run.py --polybench
```

code_benches/run.py
Others: Still don't want to merge with out an explicit check for this edge case. 

Also can you clean up the name and documentation of the "best path" function? It's a little more specific than the name let's on

code_benches/run.py
Others: 👍 something to think about in the future

code_benches/run.py
Others: Fix not needed for this PR

runtime/libc/env.c
Others: IIRC these should be put back. They are a hack for wasmception only

runtime/libc/wasi_sdk_backing.c
```diff
@@ -0,0 +1,500 @@
...
+    int fd = openat(dirfd, path, flags, mode);
+    if (fd < 0) {
+        return wasi_fromerrno(errno);
+    }
+
+    set_i32(fd_off, fd);
+    return WASI_ESUCCESS;
+}
+
+i32 wasi_unstable_fd_close(i32 fd) {
```
Others, line 326: So undefined symbols go to "env" but wasi explicitly imports from a different namespace. I didn't even realize my existing jank would handle that case 😃

code_benches/run.py
bushidocodes: I think this documentation should help with this: https://docs.python.org/3/library/argparse.html#mutual-exclusion

code_benches/run.py
```diff
@@ -105,17 +157,18 @@ def sources(self):
     # == Custom Benchmarks ==
     Program("custom_binarytrees", [16], 2 ** 14),
     Program("custom_function_pointers", [], 2 ** 14),
-    Program("custom_libjpeg", [], 2 ** 15,
-            custom_arguments=["-Wno-incompatible-library-redeclaration", "-Wno-implicit-function-declaration",
-                              "-Wno-shift-negative-value"]),
+    #Program("custom_libjpeg", [], 2 ** 15,
```
bushidocodes, line 160: I assume that this is a WASI-SDK related regression as with line 166 below? If so, I concur with @Others. My suggestion is to open an issue for each of these regressions and add TODO comments that include the relevant issue numbers.

runtime/memory/64bit_nix.c
Others: @geky I don't necessarily agree, since the spec says that `memory` is always divisible by page size. It seems our conclusion is to punt on that

Others:
Here is what needs done to merge this:

* Regarding the incompatible flags issue, we may want to move the official solution Sean suggested.
* Regarding the best path function:
> Also can you clean up the name and documentation of the "best path" function? It's a little more specific than the name let's on
* env_sin and env_cos back to wasmception specific code, as they are a hack.
* rebase off master
* add an issue to track work on non-native page sizes/the weird behavior of memory.size in the presence of non-native page sizes, and fixing up the mmap wasmception hack we currently use for non-native page sizes
* add an issue to track fixing the broken programs for wasi-sdk
* add an issue to track implementing multiple memory support (right now this may cause weirdness if we aren't asserting that we're always loading from memory 0)

Others:
* add an issue to track wasi CPU version changes

Others:
Had a go trying to rebase this, and very quickly ran into problems.
```
/Users/peachg/Projects/aWsm_rebase/wasi-sdk/bin/clang -Wl,--allow-undefined,-z,stack-size=16384,--no-threads,--stack-first,--no-entry,--export-all,--export=main,--export=dummy --target=wasm32-wasi -mcpu=mvp -nostartfiles -O3 -flto --sysroot=/Users/peachg/Projects/aWsm_rebase/wasi-sdk/share/wasi-sysroot  -O3 -flto ../dummy.c binarytrees.c -o bin/custom_binarytrees.wasm
```
Is this missing a necessary submodule?

geky:
Ah, it looks like I didn't update the instructions. You need to download wasi-sdk outside of git:

``` bash
WASI_SDK_URL=https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-8/wasi-sdk-8.0-linux.tar.gz
wget $WASI_SDK_URL -O wasi-sdk.tar.gz
mkdir -p wasi-sdk
tar xvfz wasi-sdk.tar.gz --strip-components=1 -C wasi-sdk
```

(I have this in the CI script, but I just realized it's nowhere else.)

This raises an interesting question of if the wasi-sdk should be available as a submodule. One of the benefits of moving to wasi-sdk is you don't need to compile LLVM thanks to the prebuilt binaries, but these aren't available as a git repo.

It may be valuable to also include wasi-sdk as a submodule, in which case this is the release/hash I was using for wasi-sdk:
https://github.com/WebAssembly/wasi-sdk/releases/wasi-sdk-8

Though not you would need to compile this before it would be usable, you also may need to tweak the paths to point to the compiled binaries correctly since I think the binary paths in the source differ from the prebuilt-binaries.

Others:
We can merge this, but it is totally broken on Mac.

1) Downloads a linux specific binary
2) Produces a compilation error:
```
clang -lm  -O3 -flto -g bin/custom_binarytrees.bc /Users/peachg/Projects/aWsm_rebase/runtime/runtime.c /Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c /Users/peachg/Projects/aWsm_rebase/runtime/libc/env.c /Users/peachg/Projects/aWsm_rebase/runtime/memory/cortex_m.c -o bin/custom_binarytrees_cm
warning: overriding the module target triple with x86_64-apple-macosx10.15.0 [-Woverride-module]
1 warning generated.
/Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c:280:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
}
^
/Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c:305:49: error: use of undeclared identifier 'O_RSYNC'
            ((fdflags & WASI_FDFLAG_RSYNC   ) ? O_RSYNC    : 0) |
                                                ^
/Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c:361:20: error: use of undeclared identifier 'O_RSYNC'
            ((fl & O_RSYNC)    ? WASI_FDFLAG_RSYNC      : 0) |
                   ^
/Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c:374:43: error: use of undeclared identifier 'O_RSYNC'
        ((flags & WASI_FDFLAG_RSYNC   ) ? O_RSYNC    : 0) |
                                          ^
/Users/peachg/Projects/aWsm_rebase/runtime/libc/wasi_sdk_backing.c:395:15: warning: implicit declaration of function 'fdatasync' is invalid in C99 [-Wimplicit-function-declaration]
    int res = fdatasync(fd);
```

Others:
@geky @gparmer Should we merge this and fast follow with a fix for OS-X?

Others:
I got this working on my machine with the following diff: https://gist.github.com/Others/49f7201c707edbe84dd0898ae17a185a

I think we can either apply this diff + do something smarter to pick the wasi release to download, or ship this PR, and I can follow up with a patch to fix it on mac.

PR #26, opened 2021-03-24
Resolve Mac Build Issues

bushidocodes:
A handful of fixes and QOL improvements for Mac. I made a first pass as an install script based on what I had to manually install to get things working.

I think the main thing worth discussing is the linking errors I hit (likely currently broken at head of master). I'm unclear why I didn't hit this when running the branch, but I suspect that perhaps some build asset was cached and I wasn't clear how to do the equivalent of a `make clean`. In any case, there are two functions that seem to be duplicated across a number of files. My preference would be to DRY that up in a follow-up issue, but we should make sure I understand all the different permutations of how things get compiled. It seemed like each compiled aWsm module links in both `runtime.c` and one of the memory translation units.

### DIFF
--- a/install_mac.sh
+++ b/install_mac.sh
@@ -11,6 +11,7 @@ fi
 # Install required tools
 brew install svn
 brew install wget
+brew install cmake
 
 # Install LLVM and clang stuff
 xcode-select install
--- a/install_mac.sh
+++ b/install_mac.sh
@@ -2,11 +2,10 @@
 git submodule update --init --recursive
 
 # Install brew
-if ! command -v brew &> /dev/null
-then 
-  curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh | bash -s -- -y
-else
+if [[ -x "$(command -v brew)" ]]; then
   echo "Brew install detected"
+else
+  curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh | bash -s -- -y
 fi
 
 # Install required tools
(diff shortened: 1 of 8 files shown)


runtime/libc/wasi_sdk_backing.c
```diff
@@ -371,7 +387,7 @@ i32 wasi_unstable_fd_fdstat_set_flags(i32 fd, u32 fdflags) {
         ((flags & WASI_FDFLAG_APPEND  ) ? O_APPEND   : 0) |
         ((flags & WASI_FDFLAG_DSYNC   ) ? O_DSYNC    : 0) |
         ((flags & WASI_FDFLAG_NONBLOCK) ? O_NONBLOCK : 0) |
-        ((flags & WASI_FDFLAG_RSYNC   ) ? O_RSYNC    : 0) |
+        ((flags & WASI_FDFLAG_RSYNC   ) ? O_RSYNC : 0) |
```
Others, line 390: Can you fix this indentation?

install_mac.sh
```diff
@@ -0,0 +1,36 @@
...
+if [[ -x "$(command -v rustup)" ]]; then
+  rustup update
+else
+  curl https://sh.rustup.rs -sSf | bash -s -- -y
+fi
+source "$HOME/.cargo/env"
+export PATH="$HOME/.cargo/bin:$PATH"
+
+# Build project
+cargo build --release
```
Others, line 36: No newline at end of file

Others (changes requested):
Patch to fix the instructions issue, and move env_sin/env_cos (which was lagging ): https://gist.github.com/Others/6eb4d17beedbee102315f5df09ed73f8

Can you apply that patch as part of this PR? After you do that, can you also spot check `run.py` runs cleanly through for both `wasmcpetion` and `wasi-sdk`?

Sorry to ask for more work, just want to make sure master is in a clean slate.

Others (approved):

PR #20, opened 2020-08-19
Add CI built on GitHub Actions

geky:
Note: This depends on https://github.com/gwsystems/aWsm/pull/19, but it can be disentangled if needed. See https://github.com/gwsystems/aWsm/compare/51dc1d1...geky:testing-squished to see what's changed.

This PR adds support for CI built on GitHub Actions. I will tinker with it to make sure its working, but it may just automatically run. It should take any commits/PRs against this repo, build them against both wasmception and wasi-sdk, and push the resulting benchmarks.csv as a "artifact" to the workflow. I will aadd a comment once its working.

There's a small TODO list:
1. Host a precompiled wasmception binary somewhere. Currently I have the wasmception binary as a release on my repo [here](https://github.com/geky/aWsm/releases/tag/wasmception-0.1). You should just be able to download the `wasmception-0.1.tar.gz` and reupload to the repo you want to host from. Should this be Others/wasmception or gwsystems repo?

   Once hosting in a correct place I can update this PR to point there instead of my own repo.

2. Make sure GitHub actions is enabled on gwsystems/awsm. Just need to see what's needed here. This PR may just work without GitHub-side config, or it may not.

Anyways I just wanted to get this PR up for feedback, it should hopefully help build confidence on PRs to here.

PR #27, opened 2021-03-26
chore: Cherry pick @geky diff from #20

bushidocodes:
This PR reapplies @geky's diff and modifies it to use the wasmception release tarball in our gwsystems repo. Should be good to squash merge and then this and PR #20 can be closed.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -59,6 +59,10 @@ jobs:
 
       - name: Code benches
         run: |
+          pwd
+          ls wasmception
+          ls wasmception/dist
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -0,0 +1,70 @@
+# CI workflow
+name: test
+on: [push, pull_request]
+
+env:
+  # Rust equivalent of -Werror
+  RUSTFLAGS: --deny warnings
```
Others, line 7: This can be flaky, but fine for now

Others (approved):
You are a machine!!

PR #21, opened 2020-08-19
Rename Silverfish -> aWsm

geky:
And here is the Silverfish -> aWsm rename.

This PR is just this script, applied on top of https://github.com/gwsystems/aWsm/pull/20:
```
git ls-tree HEAD --name-only -r | grep -v wasmception | xargs sed -i -e 's/silverfish/awsm/g' -e 's/Silverfish/aWsm/g' -e 's/SILVERFISH/AWSM/g'
```

I've mostly created a PR just so that everything is tracked. It's easy enough to rerun, so this can be easily rebased whenever you think it's best to bring it in.

PR #28, opened 2021-03-27
chore: run @geky one-liner from #21

bushidocodes:


README.md
```diff
@@ -25,11 +25,6 @@ Wasm provides a number of benefits outside of the web including:
 	Wasm is a high-level assembly, independent from any specific hardware.
 	This has the potential to provide a universal specification of computation that can execute in the cloud, on the edge, or in an embedded device.
 
-**A note on naming.**
```
Others, line 28: I use this on my resume. Can I ask this not be retained for selfish reasons?

runtime/libc/env.c
```diff
@@ -1,4 +1,4 @@
-/* https://github.com/gwsystems/silverfish/blob/master/runtime/libc/libc_backing.c */
+/* https://github.com/gwsystems/awsm/blob/master/runtime/libc/libc_backing.c */
```
Others, line 1: Isn’t this a historical link? Will that work?

Others:
Also is it “awsm” or “aWsm”?

README.md
gparmer: Yeah, we should retain this comment.

We can remove the "lingering references" and "patience" comment.

bushidocodes:
@gparmer, @Others - RE `aWsm` vs `awsm`, my personal preference would be to just going with `awsm`. I don't think we really lose anything semantically and we end up with a lot fewer case-related footguns. That would include renaming the repo from `aWsm` to `awsm`, which technically wouldn't affect resolution since URLs are not case sensitive, but it would mean that the repo would more consistently clone into an `awsm` directory rather than an `aWsm` directory that I am two dumb to change directory into on first try.

README.md
bushidocodes: I attempted a rewrite of the acknowledgements section to capture these contributions. Happy to credit Gregor, but I'm not clear why we need to have a note on the silverfish name after all references have been wiped, since it's largely historical at that point.

runtime/libc/env.c
bushidocodes: Doesn't resolve. No clue what the point of it is either, so I'll just delete it.

README.md
Others: I think it’s generally good practice when renaming something to leave a note of the previous name.

For example I’ve linked to this repo and mentioned the project name externally before, and I don’t really want to go track down those references and change the name I used.

Any harm in adding a “previously known as” or “Gregor worked on this under the name ...” sentence?

gparmer: No harm. We should do that.

gparmer:
I'd also prefer `awsm` rather than `aWsm`.

README.md
bushidocodes: Added a "previously known as" note under the title.

Others (approved):

PR #31, opened 2021-03-29
refactor: Clang and Cargo Auto-formatting and CI Checks

bushidocodes:


Others:
Any way this can be a separate check from the wasi/wasmception ones? Not a fan that it blocks running the tests

Others:
Please make sure this isn’t going to create a huge formatting diff. This does not need to match composite. The code is not OS code and is intentionally written in a Rust-esk style

For example, I think we are indented by 4 spaces currently, so don’t change that. Also please don’t let this move all the curly braces around

Honestly I think this is relatively low priority, since there is a very small amount of C code written by us in the project

example_code/mutual_recursion.c
```diff
@@ -9,8 +9,6 @@ int f(int i) {
 }
 
 int g(int i) {
-    if (i > 0) {
-        return f(i - 1) + 1;
-    }
+    if (i > 0) { return f(i - 1) + 1; }
```
Others, line 12: Don't think we want ifs on one line

src/wasm.rs
```diff
@@ -423,32 +444,101 @@ pub enum Instruction {
...
-    I64Load32S { flags: u32, offset: u32 },
-    I64Load32U { flags: u32, offset: u32 },
-    I64Store32 { flags: u32, offset: u32 },
-
-    F32Load { flags: u32, offset: u32 },
-    F32Store { flags: u32, offset: u32 },
-
-    F64Load { flags: u32, offset: u32 },
-    F64Store { flags: u32, offset: u32 },
+    I32Load {
```
Others, line 447: I find the new formatting here a bit worse. Any way to exclude this enum?

Others (commented):
Two nits

example_code/mutual_recursion.c
bushidocodes: Should be fixed!

src/wasm.rs
bushidocodes: Should be fixed!

Others:
Beautiful!

bushidocodes:
@Others - I changed the formatter configs for your nits and reran our format script. Please approve if the source reflects your wishes. I'll wait to merge until the code benches run to completion, but it's very unlikely that they'll fail.

Others (approved):
Shipit

PR #33, opened 2021-07-05
Implement if-else instruction

Others:
This is completly untested (didn't even re-run the test suite). It
does build though!

I personally garuntee at least one subtle control flow bug in this
patch--probably needs a test before merging.

EDIT: This now has tests / is tested! Thanks Sean!


gparmer:
@Others Note the test that @bushidocodes added in https://github.com/gwsystems/aWsm/pull/35 This currently fails, but that's not surprising for a WiP.

gparmer:
Had a look through this, and I don't have any useful feedback. I very much appreciate how much you're able to use much of the existing infrastructure to generate this feature. If, indeed, the wasm file for if/else that @bushidocodes causes this to blow up, I don't immediately see where the error is coming from.

Let me know if you need more directed feedback. I'll need to spend a lot of time diving in to the logic, in that case.

src/codegen/block.rs
```diff
@@ -235,6 +249,9 @@ pub fn compile_block<'a, 'b>(
                 b.build_br(tt.bb);
                 // Fetch the ctx
                 let if_ctx_resolved = if_ctx.as_ref().expect("malformed wasm -- else with no if");
+                // Ensure we don't build a duplicate else
+                assert_eq!(if_ctx_resolved.else_built.get(), false);
```
gparmer, line 253: Can we spit out a useful error message here even if we do just end up bombing out?

src/codegen/block.rs
```diff
@@ -194,20 +197,31 @@ pub fn compile_block<'a, 'b>(
...
+                        else_built: &else_built
                     }),
                     if_branch_locals,
                     if_branch_block,
                     remaining_instructions,
                 );
 
+                // Handle the case where there is no else
+                if !else_built.get() {
+                    let fake_stack = Vec::new(); // Tacitly assumes an if with no else can't produce a value
```
gparmer, line 218: @bushidocodes Just wanted to confirm that this assumption is correct, and you just learned a bunch of wasm ;-) Does this assumption make sense? Also, if @Others doesn't need the sanity check here, I'm happy with it.

gparmer:
Brilliant guys! Added a suggestion for better feedback for incorrect wat files, and just want to validate an assumption. Other than that, this looks quite good to go!

Thanks a ton for this @Others !

src/codegen/block.rs
bushidocodes: [Re-posting from Slack]. `if` expects and consumes a single argument/operand of type i32 which is evaluated as a bool (0 is false. everything else is true). This is logically popped from the stack and consumed. All conditional logic is terminated by either an `else` or an `end` , both of which expect no arguments/operands. That means that the instructions in the unit of execution following an if or an else can add to the stack, but when this unit of execution terminates (as indicated by an `else` or an `end`), the stack must be in the same state as when it started. So the net impact of running a complex if/else is that a single value evaluated as a boolean for the conditional is popped from the stack. Stated differently, I think that means that "an if of any kind cannot produce a value"

bushidocodes (approved):

src/codegen/block.rs
```diff
@@ -194,20 +197,39 @@ pub fn compile_block<'a, 'b>(
...
+                // Handle the case where there is no else
+                if !else_built.get() {
+                    // Tacitly assumes an if with no else can't produce a value
+                    // This is safe, since WASM assumes you always statically know the type of everything on the stack.
+                    // Consider the following scenario:
+                    // - We have an u32 on the stack
+                    // - We execute an `if` with no else, that produces a u64
+                    // - Now the stack has a u32 on it, and MAYBE a u64
+                    // - What is the type of value yielded by the next pop instruction?
+                    // This is the kind of issue that wasm avoids by dissallowing this scenario
```
gparmer, line 225: Brilliant comment, thanks!

gparmer: dissallowing -> disallowing

gparmer (approved):
Sorry for the delay. I missed the additions to this.

PR #34, opened 2021-07-07
Makefile

bushidocodes:
An example Makefile provided to help provide an example for external researchers


emil916 (commented):
Looks good. The only thing is, I noticed the `SUPPORTS_MPX` variable is never used. Should there be a particular make command for **mpx** where that variable may be used?

bushidocodes:
> Looks good. The only thing is, I noticed the `SUPPORTS_MPX` variable is never used. Should there be a particular make command for **mpx** where that variable may be used?

Good catch. I removed those rules in a previous commit since both segmentation and MPX are effectively an architectural curiosity and not really useful anymore. Will remove this zombie variable!

code_benches/example.mk
```diff
@@ -0,0 +1,123 @@
...
+# This project currently uses a hand-spun build system and test running written (run.py)
+# The intent behind this sample makefile is to illustrate how one might use awsm using
+# a standard build tool.
+
+CC=clang
+OPTFLAGS=-O3 -flto
+UNAME:=$(shell uname -m)
+
+# Assuming to be called from code_benches directory
+ROOT_PATH:=$(shell cd .. && realpath .)
```
gparmer, line 12: I learn so many things from your unix hackery ;-)

code_benches/example.mk
```diff
@@ -0,0 +1,123 @@
...
+endif
+	@mkdir -p ./app_pid/bin
+	@${WASM_CC} ${WASM_FLAGS} ${WASM_LINKER_FLAGS} -Wl,-z,stack-size=256  -Wall -std=c++11 ./dummy.cpp ./app_pid/*.cpp -o ./app_pid/bin/app_pid.wasm
+
+.PHONY: app_pid.bc
+app_pid.bc: app_pid.wasm
+	@mkdir -p ./app_pid/bin
+	@${AWSM_CC} ./app_pid/bin/app_pid.wasm -o ./app_pid/bin/app_pid.bc
+	
+# Generate bitcode using the "fast unsafe implementations" option
```
gparmer, line 75: Likely good to be explicit in the comment:

`Generate LLVM bitcode...`

code_benches/example.mk
```diff
@@ -0,0 +1,123 @@
...
+
+# TODO: Directly executing the native binary does not write output to the console as with the
+# various awsm builds. This behavior is observable when building with run.py as well.
+.PHONY: app_pid_native
+app_pid_native:
+	@mkdir -p ./app_pid/bin
+	@${CC} -lm ${OPTFLAGS} -Wall -std=c++11 ./app_pid/*.cpp -o ./app_pid/bin/app_pid_native
+
+.PHONY: app_pid.wasm
+app_pid.wasm:
```
gparmer, line 61: Should have dependencies on the *.cpp files here, I think.

gparmer (approved):
Looks great. I think that the missing dependencies on *.cpp/*.c files is the only thing that doesn't seem complete to me. Easy to fix, so we don't need to go another round ;-) Approved.

PR #38, opened 2021-07-14
feat: WASI-SDK 12 implementation

bushidocodes:
This PR is my attempt at upgrading to WASI-SDK 12 and updating the runtime to use the latest WASI ABI. It stubs out WASI syscalls that we have not yet implemented.


bushidocodes:
Just saw the tests. I suspect that CI is breaking because it is using WASI-SDK 8 and needs to be updated.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -339,7 +334,45 @@ i32 wasi_unstable_path_open(
...
+ * @param argv_p_offset 
+ * @param argv_buf_offset 
+ * @return i32 
+ */
+i32 wasi_snapshot_preview1_args_get(i32 argv_p_offset, i32 argv_buf_offset){
+    char *argv_buf = get_memory_ptr_for_runtime(argv_buf_offset, runtime_argv_buffer_len);
+    strncpy(argv_buf, runtime_argv_buffer, runtime_argv_buffer_len);
+
+    u32 *argv_p = (u32 *)get_memory_ptr_for_runtime(argv_p_offset, sizeof(i32));
+    *argv_p = argv_buf_offset;
```
gparmer, line 350: Strange API that passes in a chunk of memory and expects a pointer to it back. Seems odd.

https://docs.rs/wasi/0.9.0+wasi-snapshot-preview1/wasi/fn.args_get.html confirms your intuition. I might put `args_sizes_get` before this function to mimic the dependency.

runtime/runtime.c
```diff
@@ -253,9 +253,35 @@ void* allocate_n_bytes_ptr(u32 n) {
...
-IMPORT i32 wasmf_main(i32 a, i32 b);
+IMPORT void wasmf__start(void);
+
+int runtime_argc = 0;
+
+char *runtime_argv_buffer = NULL;
+int runtime_argv_buffer_len = 0;
+
+void runtime_cleanup() {
+    // Cancel any pending timeout
```
gparmer, line 264: Remove the commented out code?

runtime/runtime.c
```diff
@@ -253,9 +253,35 @@ void* allocate_n_bytes_ptr(u32 n) {
...
+    runtime_argv_buffer_len = 0;
+}
 
 int runtime_main(int argc, char** argv) {
+    // Set argc and argv to globals, these are later used by the WASI syscalls 
+    runtime_argc = argc;
+
+    // Copy the null terminated args one after the other into a global with the format WASI calls expect
+    for (int i = 0; i < argc; i++) {
+        size_t arg_len = strlen(argv[i]);
```
gparmer, line 279: Is this loop not copying in the `\0`s? If you print out the `runtime_argv_buffer`, do you see all arguments (wrong), or only the first (correct)?

Or are the arguments supposed to be space-separated arguments (in which case we should add a ` ` before the next string).

gparmer (changes requested):
Validate that the initial retrieval (in `c`) of the arguments is correct. Other than that, only small updates.

runtime/runtime.c
bushidocodes: I was originally copying in the null terminators, but I think that's actually wrong. Based on what I see running my wat sample under wasmtime, the backing buffer seems to not be delimited at all, and the vector of base addresses returned by `args_sizes_get` seems to be how it's chopped up. I'll sanity check that with a C example though!

code_benches/run.py
```diff
@@ -117,7 +117,7 @@ def getmtime_or_zero(x):
 ])
 
 # These flags are all somewhat important -- see @Others for more information
-WASM_LINKER_FLAGS = "-Wl,--allow-undefined,-z,stack-size={stack_size},--no-threads,--stack-first,--no-entry,--export-all,--export=main,--export=dummy"
```
Others, line 120: Can you explain the flag removal here?

runtime/runtime.c
```diff
@@ -253,9 +253,41 @@ void* allocate_n_bytes_ptr(u32 n) {
 WEAK void populate_globals() {}
 
 // Code that actually runs the wasm code
-IMPORT i32 wasmf_main(i32 a, i32 b);
+IMPORT void wasmf__start(void);
```
Others, line 256: This diff makes us super wasi specific--prefer if you push this decision into the platform specific file

runtime/runtime.c
bushidocodes: This is previously commented out code that Gregor wrote. I just shifted the cleanup part here to make sure that it runs via atexit. I figure we should defer figuring out the utility of this code until later, so keep it commented out for now.

runtime/runtime.c
```diff
@@ -253,9 +253,41 @@ void* allocate_n_bytes_ptr(u32 n) {
...
+
```
Others, line 257: Overall feedback: you've taken apart the abstraction layer here (that strives to make this platform agnostic, and support wasmception and wasi-sdk)

I think it's not too much work to instead preserve this and instead improve the abstraction. YMMV

runtime/runtime.c
Others: Do we want timeout support in aWsm?

code_benches/run.py
bushidocodes: Mostly. The flags `--no-entry,--export-all,--export=main,--export=dummy` were previously required to make sure that the various symbols were available to the runtime. Most of this logic is now handled by WASI ABT via `_start`. Removing the exports encapsulates the module. I don't really have an explanation for `--stack-first` beyond it didn't seem needed.

gparmer: Do you have any idea what `--stack-first` is? Or does @Others ? I'd imagine it is data stack placement. Pretty cool we can control that.

runtime/runtime.c
gparmer: I can't imagine we would in runtime.c. Maintaining time/using setjmp etc... are all platform specific. Thoughts?

Others: I defer to Gabe on this one, it was just an experiment, and we can always revive it later

code_benches/run.py
Others: I think we want `--stack-first` I believe it means the stack overflows to a negative offset (or rather underflows a u32 offset), rather than to valid memory addresses in linear memory. 

That's just what I vaguely remember, probably worth actually verifying that. (I vaguely also think it was destined to become the default.)

Also the `dummy` thing is a hack I've lost full context for (see the `dummy` function in the runtime). I think it may still be needed for wasmception? Not sure

runtime/runtime.c
bushidocodes: Yeah, I can't think of a specific target use case. Since we'll have it in the commit history, I'll remove the commented out code.

example_code/args/.gitignore
```diff
@@ -0,0 +1,3 @@
+args_vm
+args.bc
+args.wasm
```
Others, line 3: Nit: Newline at end of file

example_code/args/Makefile
```diff
@@ -0,0 +1,12 @@
...
+	rm -f args.wasm args.bc args_vm
+
+args.wasm: args.c
+	../../wasi-sdk/bin/clang -Wl,--allow-undefined,-z,stack-size=32768,--threads=1 --target=wasm32-wasi -mcpu=mvp -O3 -flto --sysroot=../../wasi-sdk/share/wasi-sysroot  -O3 -flto args.c -o args.wasm
+
+args.bc: args.wasm
+	../../target/release/awsm  args.wasm -o args.bc
+
+args_vm: args.bc ../../runtime/runtime.c ../../runtime/libc/wasi_sdk_backing.c ../../runtime/libc/env.c ../../runtime/memory/64bit_nix.c
+	clang -lm  -O3 -flto -g args.bc ../../runtime/runtime.c ../../runtime/libc/wasi_sdk_backing.c ../../runtime/libc/env.c ../../runtime/memory/64bit_nix.c -o args_vm
```
Others, line 12: Nit: Newline at end of file

example_code/args/args.c
```diff
@@ -0,0 +1,9 @@
+#include <stdio.h>
+
+int main(int argc, char **argv)
+{
+    for (int i = 0; i < argc; i++)
+    {
+        printf("Arg %d: %s\n", i, argv[i]);
+    }
+}
```
Others, line 9: Nit: Newline at end of file

runtime/libc/cortex_m_backing.c
```diff
@@ -4,6 +4,41 @@ int printf_(const char* format, ...);
...
+int runtime_main(int argc, char** argv) {
+    // Setup the linear memory and function table
+    alloc_linear_memory();
+    populate_table();
+    // Setup our allocation logic
+    runtime_heap_base = wasmg___heap_base;
+    printf("starting rhb %d\n", runtime_heap_base);
+    if (runtime_heap_base == 0) {
+        runtime_heap_base = memory_size;
+    }
```
Others, line 16: Is this how CM worked before? Jesus

How does this even work before `populate_globals` runs?

runtime/libc/cortex_m_backing.c
```diff
@@ -4,6 +4,41 @@ int printf_(const char* format, ...);
...
+    // Setup the global values (if needed), and populate the linear memory
+    switch_out_of_runtime();
+    populate_globals();
+    switch_into_runtime();
+    populate_memory();
```
Others, line 21: Let's create a function for this. I think if theory 
```
    // Setup the linear memory and function table
    alloc_linear_memory();
    populate_table();
    switch_out_of_runtime();
    populate_globals();
    switch_into_runtime();
    populate_memory();
```
Could all run at the start of the function

runtime/libc/wasi_sdk_backing.c
```diff
@@ -27,21 +27,63 @@
...
+    free(runtime_argv_buffer_offsets);
 }
 
-// The symbol the binary gives us to init libc
-extern void wasmf___wasm_call_ctors(void);
+int main(int argc, char* argv[]) {
+    // Set argc and argv to globals, these are later used by the WASI syscalls 
+    runtime_argc = argc;
+
+    runtime_argv_buffer_offsets = calloc(argc, sizeof(u32));
```
Others, line 47: Use a global variable, not calloc. Don't want to allocate in the runtime generally

runtime/libc/wasi_sdk_backing.c
```diff
@@ -27,21 +27,63 @@
...
+
+    // Copy the null terminated args one after the other into a global with the format WASI calls expect
+    for (int i = 0; i < argc; i++) {
+        runtime_argv_buffer_offsets[i] = runtime_argv_buffer_len;
+        size_t arg_len = strlen(argv[i]) + 1;
+        runtime_argv_buffer = realloc(runtime_argv_buffer, runtime_argv_buffer_len + arg_len);
```
Others, line 53: Also would be nice to avoid allocating here

runtime/libc/wasi_sdk_backing.c
```diff
@@ -476,41 +567,88 @@ i32 wasi_unstable_path_filestat_get(i32 fd, u32 flags, u32 path_off, u32 path_le
...
+
+// TODO: wasi_snapshot_preview1_random_get
+
+// TODO: wasi_snapshot_preview1_sched_yield
+
+// TODO: wasi_snapshot_preview1_sock_recv
+
+// TODO: wasi_snapshot_preview1_sock_send
+
+// TODO: wasi_snapshot_preview1_sock_shutdown
```
Others, line 654: Nit: No newline at end of file

runtime/runtime.c
```diff
@@ -226,84 +224,5 @@ void* allocate_n_bytes_ptr(u32 n) {
...
-//    signal(SIGALRM, handle_sigalarm);
-//    ualarm(wasm_execution_timeout_ms * 1000, 0);
-//}
-//
-//void cancel_timeout() {
-//    ualarm(0, 0);
-//}
-
 // If we are using runtime globals, we need to populate them
 WEAK void populate_globals() {}
```
Others, line 228: OT: Runtime globals turned out to be a garbage feature, I legitimately can think of no use for them. We should make an issue to review removing them

Others (changes requested):
Mostly just nitpicks. Good job on this

runtime/libc/cortex_m_backing.c
bushidocodes: Not sure! I haven't yet gotten familiar with `populate_globals`, and a cursory glance at `insert_runtime_globals` goes over my head. 🤣😅 Before this PR, the entry point of all three variants' (wasmception, wasi-sdk, and cortex-m) called `runtime_main`, which had this same order of instructions. So all of the variants appear to have worked this way, not just cortex m. I think that this was logged via a printf like `printf("starting rhb %d\n", runtime_heap_base);`. I think I've only see this log out `starting rhb 0`. Not sure what that means.

bushidocodes: Based on your comment, I'll move this into the `runtime_init` function after the calls to populate the table, globals, and memory. Sounds like there might be more to address here later.

runtime/runtime.c
bushidocodes: Opened as #40 !

code_benches/run.py
```diff
@@ -86,18 +86,20 @@ def getmtime_or_zero(x):
...
 WASMCEPTION_BACKING = "wasmception_backing.c"
+WASMCEPTION_LINKER_FLAGS = "-Wl,--allow-undefined,-z,stack-size={stack_size},--no-threads,--stack-first,--no-entry,--export-all,--export=main,--export=dummy"
 
 WASI_SDK_PATH = ROOT_PATH + "/wasi-sdk"
 WASI_SDK_CLANG = WASI_SDK_PATH + "/bin/clang"
 WASI_SDK_SYSROOT = WASI_SDK_PATH + "/share/wasi-sysroot"
-WASI_SDK_FLAGS = "--target=wasm32-wasi -mcpu=mvp -nostartfiles -O3 -flto"
+WASI_SDK_FLAGS = "--target=wasm32-wasi -mcpu=mvp -O3 -flto"
 WASI_SDK_BACKING = "wasi_sdk_backing.c"
+WASI_SDK_LINKER_FLAGS = "-Wl,--allow-undefined,-z,stack-size={stack_size},--threads=1,--stack-first"
```
gparmer, line 96: Brilliant, thanks for this addition.

runtime/libc/cortex_m_backing.c
```diff
@@ -4,6 +4,28 @@ int printf_(const char* format, ...);
 
 volatile int CORTEX_M_ARG_C = 0;
 
+int runtime_main(int argc, char** argv) {
```
gparmer, line 7: This all makes sense to me.

runtime/libc/wasi_sdk_backing.c
gparmer: I don't think that's possible here as this is allocating the array of arguments, and we don't statically know how many arguments there are.

runtime/libc/wasi_sdk_backing.c
gparmer: @Others: Is there a way to avoid the dynamic allocation, though? The maximum length of the argument array is pretty stupidly large. Best we could do here (unless I'm mistaken) is to iterate through arguments, accumulating `strlen`, and do a *single* allocation instead of the repetitive `realloc`. That would be more memory and runtime effiicent.

I'm sorry @bushidocodes , but I think we do want to do this (I should have though about this when I first saw the code). We need this in sledge (to optimize for sandbox creation time), so we might as well do it right here, and replicate there. I'm OK with this is an issue TODO if we need to get this PR accepted quickly.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -27,22 +27,55 @@
...
+u32 runtime_argc = 0;
+char *runtime_argv_buffer = NULL;
+u32 runtime_argv_buffer_len = 0;
+u32 *runtime_argv_buffer_offsets = NULL;
+
+void runtime_cleanup() {
+    // TODO: What if atexit triggers when inside of runtime?
+    // Before implementing more complex state machine transitions, we should decide if segmentation
+    // (the only thing that uses these calls) merits this additional complexity.
+    switch_into_runtime();
```
gparmer, line 41: Shouldn't we already be in runtime? This is called after `main` returns.

I don't have strong opinions on if we should keep it or not.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -27,22 +27,55 @@
...
 
-// Called before wasmf_main
-void stub_init() {
     switch_out_of_runtime();
-    wasmf___wasm_call_ctors();
+    wasmf__start();
     switch_into_runtime();
+    
+    // TODO: Improve "implicit exit" path
+    // My understanding is that _start can optionally return. How can we get this?
```
gparmer, line 75: I'm not sure this is worth handling. If our constructors fail...meh?

runtime/runtime.c
```diff
@@ -226,84 +224,20 @@ void* allocate_n_bytes_ptr(u32 n) {
...
-    u32 array_offset = allocate_n_bytes(argc * sizeof(i32));
-    u32* array_ptr = get_memory_ptr_void(array_offset, argc * sizeof(i32));
-    for (int i = 0; i < argc; i++) {
-        size_t str_size = strlen(argv[i]) + 1;
-        u32 str_offset = allocate_n_bytes(str_size);
-        char* str_ptr = get_memory_ptr_for_runtime(str_offset, str_size);
-        strcpy(str_ptr, argv[i]);
-        array_ptr[i] = str_offset;
+    runtime_heap_base = wasmg___heap_base;
+    printf("starting rhb %d\n", runtime_heap_base);
```
gparmer, line 239: Can we remove the prints from this so that we can use it more generically? We don't want these prints in sledge, for example.

gparmer (approved):
Up to you if you do the PR, and add an issue to update the realloc stuff, or if you want to update it now. Given this option, I'm approving.

runtime/runtime.c
bushidocodes: I added this back in because Gregor mentioned that he was surprised that this was previously occurring before a function call to populate the globals. I've only ever seen this log 0, so I'm not sure if this is working as expected or not.

runtime/libc/cortex_m_backing.c
```diff
@@ -4,6 +4,28 @@ int printf_(const char* format, ...);
...
+	runtime_init();
```
Others, line 8: indentation!

runtime/libc/wasi_sdk_backing.c
```diff
@@ -27,188 +28,95 @@
...
+    runtime_args_init(argc, argv);
 
-enum wasi_fdflags {
-    WASI_FDFLAG_APPEND      = 0x0001,
-    WASI_FDFLAG_DSYNC       = 0x0002,
-    WASI_FDFLAG_NONBLOCK    = 0x0004,
-    WASI_FDFLAG_RSYNC       = 0x0008,
-    WASI_FDFLAG_SYNC        = 0x0010,
-};
+    atexit(runtime_cleanup);
```
Others, line 102: I find the multiple `atexit` calls somewhat confusing

runtime/runtime.c
```diff
@@ -226,84 +224,19 @@ void* allocate_n_bytes_ptr(u32 n) {
...
-    for (int i = 0; i < argc; i++) {
-        size_t str_size = strlen(argv[i]) + 1;
-        u32 str_offset = allocate_n_bytes(str_size);
-        char* str_ptr = get_memory_ptr_for_runtime(str_offset, str_size);
-        strcpy(str_ptr, argv[i]);
-        array_ptr[i] = str_offset;
+    runtime_heap_base = wasmg___heap_base;
+    if (runtime_heap_base == 0) {
+        runtime_heap_base = memory_size;
     }
```
Others, line 241: This seems... sketch

runtime/runtime.h
```diff
@@ -153,3 +153,7 @@ void stub_init();
 
 // The runtime entrypoint must be called
 int runtime_main(int argc, char** argv);
+
+extern WEAK u32 wasmg___heap_base;
+extern u32 runtime_heap_base;
+extern void runtime_init();
```
Others, line 159: Why `extern` and not just a prototype?

runtime/runtime.h
```diff
@@ -153,3 +153,7 @@ void stub_init();
 
 // The runtime entrypoint must be called
 int runtime_main(int argc, char** argv);
+
+extern WEAK u32 wasmg___heap_base;
+extern u32 runtime_heap_base;
```
Others, line 158: docs why we need these here pls

Others (changes requested):
Haven't reviewed all the WASI syscalls yet

runtime/runtime.c
gparmer: Actionable feedback here?

Others: You're right, this was un-actionable. Just looking for a comment in the code on why we want to assume the heap base is `memory_size` but only sometimes.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -181,6 +181,8 @@ enum wasi_rights {
     WASI_RIGHT_SOCK_SHUTDOWN            = 0x0000000010000000ULL,
 };
 
+typedef uint16_t wasi_errno_t;
```
gparmer: OMG, I can't thank you enough for fixing this up. Was a pain in the ass to cross reference all of these to the standard. Thank you!

runtime/libc/wasi_sdk_backing.c
```diff
@@ -351,74 +200,189 @@ static wasi_errno_t wasi_fromerrno(int errno_) {
...
+ * @return WASI_ESUCCESS
  */
-wasi_errno_t wasi_snapshot_preview1_args_get(u32 argv_p_offset, u32 argv_buf_offset){
-    char *argv_buf = get_memory_ptr_for_runtime(argv_buf_offset, runtime_argv_buffer_len);
+wasi_errno_t wasi_snapshot_preview1_args_get(
+    wasi_size_t argv_retptr, 
+    wasi_size_t argv_buf_retptr
+){
+    wasi_size_t *argv = (wasi_size_t *)get_memory_ptr_for_runtime(argv_retptr, sizeof(wasi_size_t) * runtime_argc);
+    char *argv_buf = get_memory_ptr_for_runtime(argv_buf_retptr, runtime_argv_buffer_len);
```
gparmer, line 224: Maybe add an issue for the future? Options include 1. returning an error (likely reasonable), 2. bombing out either by panicing, or using longjmp to return from the module.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -648,21 +1052,135 @@ wasi_errno_t wasi_snapshot_preview1_path_unlink_file(i32 fd, u32 path_off, u32 p
...
+ * @param fd
+ * @param how Which channels on the socket to shut down.
+ * @return status code
+ */
+wasi_errno_t wasi_snapshot_preview1_sock_shutdown(
+    wasi_fd_t fd,
+    wasi_sdflags_t how
+) {
+    wasi_unsupported_syscall(__func__);
+}
```
gparmer: All of this stubbed out API is also a ton. Thanks

runtime/libc/wasi_sdk_backing.c
gparmer: hmmmm, did this comment become a "dangling comment". Was referencing the TODO below. Sorry, ignore this.

gparmer (approved):
I don't see any big pending issues with this. This is just brilliant! All of the work to make this more WASI conformant is great.

runtime/libc/wasi_sdk_backing.c
bushidocodes: I'll add this to an issue for improved bound checking behavior.

bushidocodes: I rephrased this issue a bit after rereading the section on "traps" in the WebAssembly documentation. Given how this is phrased, I think setjmp/longjmp into a stubbed trap handler is probably the most future-proof solution. https://github.com/gwsystems/aWsm/issues/43

runtime/libc/wasi_sdk_backing.c
Others: It's fine as is, but it also seems we could put the buffer freeing code in "runtime_cleanup"

Not a blocker for PR

runtime/runtime.c
bushidocodes: I honestly have no clue what this is supposed to do. This was in `runtime_main` before any of my work, and all I've done is move it into `runtime_init` since the `runtime_main` logic now differs between wasmception and WASI-SDK. 

What is currently in mainline:
image


It sounds like this should be a follow up issue since none of us really know why this is here. I opening the following for this: https://github.com/gwsystems/aWsm/issues/44. Since this isn't a change from master, I'm marking this as resolved.

Others: Ack. 


I do think it maybe should always be at the end of memory (not at the heap base), since otherwise malloc might overwrite this. Honestly this is basically just a wasmception hack since wasi does APIs in a more sane way. But since you didn't modify this code it isn't at all in scope for your work

runtime/libc/wasi_sdk_backing.c
bushidocodes: For error handling associated with globals, I prefer the multiple atexit handlers because they are effectively pushed to a stack. That means if there is some sort of fault within `runtime_init`, only the dynamic allocations that succeeded end up getting cleaned up. It believe this also mirrors the way GOTOs and labels are used for local error handling in the Composite style guide.

I renamed `runtime_cleanup` to `runtime_on_module_exit` to clarify that this cb is mostly intended for debugging / logging, not cleanup of runtime globals. I also moved `switch_into_runtime` into the `proc_exit` syscall to reduce the atexit calls a bit.

runtime/runtime.h
bushidocodes: After taking another look, I don't think they were needed. Removed.

runtime/runtime.h
bushidocodes: Changed to prototype

runtime/runtime.c
bushidocodes: Thanks for the feedback! Adding your comment to the follow up issue.

.github/workflows/main.yaml
```diff
@@ -6,10 +6,10 @@ env:
   # Rust equivalent of -Werror
   RUSTFLAGS: --deny warnings
   # used during LLVM installation
-  LLVM_VERSION: 9
+  LLVM_VERSION: 12
```
Others, line 9: Looks good!

Others (approved):

PR #41, opened 2021-07-20
feat: Implement environment syscalls

bushidocodes:
* Implements the `environ_sizes_get` and `environ_get` syscalls.
* Provides a test program in C that prints the environment variables
* Documents the use of persistent state across syscalls and adds error handling for undefined use of a X_get syscall before a X_size_get syscall


example_code/environ/environ.c
```diff
@@ -0,0 +1,16 @@
...
+{
+    char **env_var = environ;
+
+    while (*(env_var) != NULL)
+    {
+        fprintf(stderr, "%s\n", *(env_var++));
+    };
+
+    return 0;
+}
```
Others, line 16: no newline

runtime/libc/wasi_sdk_backing.c
```diff
@@ -297,14 +299,26 @@ wasi_errno_t wasi_snapshot_preview1_clock_time_get(
...
+    wasi_size_t *environ_offset_vector = (wasi_size_t *)get_memory_ptr_for_runtime(environ_baseretptr, runtime_environc);
+    char *environ_buf = get_memory_ptr_for_runtime(environ_buf_baseretptr, runtime_environv_buf_len);
+    extern char **environ;
+
+    wasi_size_t environ_offset = 0;
+    for (wasi_size_t i = 0; i < runtime_environc; i++){
+        environ_offset_vector[i] = environ_buf_baseretptr + environ_offset;
+        strncpy(&environ_buf[environ_offset],environ[i], runtime_environv_buf_len - environ_offset);
+        environ_offset += (strlen(environ[i]) + 1);
+    }
```
Others, line 319: When you call `get_memory_ptr_for_runtime(environ_baseretptr, runtime_environc);` it bounds checks `runtime_environc` bytes

But then here:
```
environ_offset_vector[i] = ...
```
The `environ_offset_vector` has the type of `wasi_size_t *`, and `i` can go to `runtime_environc - 1`. So you can access `environ_baseretptr + runtime_environc * sizeof(wasi_size_t)` which is often bigger than `environ_baseretptr + runtime_environc` which is what you bounds checked.

runtime/libc/wasi_sdk_backing.c
```diff
@@ -297,14 +299,26 @@ wasi_errno_t wasi_snapshot_preview1_clock_time_get(
...
 wasi_errno_t wasi_snapshot_preview1_environ_get(
-    wasi_size_t environ_retptr, 
-    wasi_size_t environ_buf_retptr
+    wasi_size_t environ_baseretptr, 
+    wasi_size_t environ_buf_baseretptr
 ) {
-    wasi_unsupported_syscall(__func__);
+    wasi_size_t *environ_offset_vector = (wasi_size_t *)get_memory_ptr_for_runtime(environ_baseretptr, runtime_environc);
+    char *environ_buf = get_memory_ptr_for_runtime(environ_buf_baseretptr, runtime_environv_buf_len);
+    extern char **environ;
```
Others, line 312: Why do we need an extern here?

runtime/libc/wasi_sdk_backing.c
```diff
@@ -297,14 +299,26 @@ wasi_errno_t wasi_snapshot_preview1_clock_time_get(
...
+
+    wasi_size_t environ_offset = 0;
+    for (wasi_size_t i = 0; i < runtime_environc; i++){
+        environ_offset_vector[i] = environ_buf_baseretptr + environ_offset;
+        strncpy(&environ_buf[environ_offset],environ[i], runtime_environv_buf_len - environ_offset);
```
Others, line 317: Why not just copy the whole vector at once?

runtime/libc/wasi_sdk_backing.c
```diff
@@ -318,7 +332,17 @@ wasi_errno_t wasi_snapshot_preview1_environ_sizes_get(
     wasi_size_t environc_retptr, 
     wasi_size_t environv_buf_len_retptr
 ) {
-    wasi_unsupported_syscall(__func__);
+    extern char **environ;
+
+    for (char **cursor = environ; *cursor != NULL; cursor++) {
+        runtime_environc++;
+        runtime_environv_buf_len += (strlen(*cursor) + 1);
```
Others, line 339: Why are you modifying globals in what should be a stateless function?

Others (changes requested):

runtime/libc/wasi_sdk_backing.c
bushidocodes: It's part of the API. From `environ.7`: "This variable must be declared in the user program, but is declared in the header file <unistd.h> if the _GNU_SOURCE feature test macro is defined."

runtime/libc/wasi_sdk_backing.c
bushidocodes: The `WASI` syscall `environ_get` is strange because it long passes the base address of a buffer and not the length. According to the spec, the logic for this is `The sizes of the buffers should match that returned by environ_sizes_get.` That means that a WASI program must call `environ_args_get` before `environ_get`. Because of this dependency, I'm using globals to avoid having to re-execute the logic of `environ_sizes_get` in `environ_get`. If this bugs you, I could potentially create a utility function that memoizes the stat of runtime_environc and runtime_environv_buf_len into static locals.

runtime/libc/wasi_sdk_backing.c
bushidocodes: There's an extra bit of indirection. `environ` points to an array of pointers to strings, so they not guaranteed to consecutive.

PR #42, opened 2021-07-21
feat: Implement random_get syscall

bushidocodes:
- Implements the `random_get` syscall
- There isn't a clear libc mapping to this, so wrote a WAT test program that calls the syscall direct and prints the random values to STDOUT

### DIFF
--- a/runtime/libc/wasi_sdk_backing.c
+++ b/runtime/libc/wasi_sdk_backing.c
@@ -5,6 +5,7 @@
 #include <printf.h>
 #include <setjmp.h>
 #include <signal.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


runtime/libc/wasi_sdk_backing.c
```diff
@@ -1104,7 +1104,16 @@ wasi_errno_t wasi_snapshot_preview1_random_get(
     wasi_size_t buf_baseretptr,
     wasi_size_t buf_len
 ) {
-    wasi_unsupported_syscall(__func__);
+    srandom(time(NULL));
```
Others, line 1107: Bad practice--we should do an if and initialize using `/dev/urandom`

Others (changes requested):

runtime/libc/wasi_sdk_backing.c
bushidocodes: I added support for `/dev/urandom` and kept the existing implementation in place as a fallback for more minimal POSIX environments. I modified the behavior to only seed on the first invocation.

PR #47, opened 2021-07-30
uvwasi

bushidocodes:
This is a WIP PR that attempts to get us to parity with the WASI support offered by Node.js by using uvwasi. 

\ (@^0^@) /

An initial pass of the syscalls is complete in `runtime/libc/uvwasi_backing.c`. The key exception is the `socket_` calls because these are tentatively planned to be removed in favor of generic read and write interfaces. These assert out, which also is what Node.js does with these calls.

@gparmer verbally brought up a concern with my use of VLAs on the stack, which causes some security issues. I agree, but would like to defer that to a follow up issue and PR.

Glancing through all these files, I think the test organization is pretty messy as well! I suggest that I clean this up in a follow up effort.

I've pulled in the Node.js test suite for WASI, and some of the tests seems to fail. I've got additional work cleaning up these tests and fixing whatever is failing. I consider passing all these tests my minimum baseline for merging this PR. Not there yet, but wanted to get this up in case anyone has cycles to review what I have so far over the weekend.


example_code/args/args.c
```diff
@@ -1,9 +1,12 @@
...
+#include <stdlib.h>
 
 int main(int argc, char **argv)
 {
+    fprintf(stderr, "Argc: %d\n", argc);
     for (int i = 0; i < argc; i++)
     {
         printf("Arg %d: %s\n", i, argv[i]);
     }
+    exit(EXIT_SUCCESS);
```
Others, line 11: Is this different from a return? Or is this just testing exit?

runtime/memory/64bit_nix.c
```diff
@@ -50,9 +50,13 @@ i32 instruction_memory_grow(i32 count) {
     return prev_size;
 }
 
-INLINE char* get_memory_ptr_for_runtime(u32 offset, u32 bounds_check) {
+INLINE void check_bounds(u32 offset, u32 bounds_check) {
     // Due to how we setup memory for x86, the virtual memory mechanism will catch the error, if bounds < WASM_PAGE_SIZE
     assert(bounds_check < WASM_PAGE_SIZE || (memory_size > bounds_check && offset <= memory_size - bounds_check));
```
Others, line 55: OT: this is probably not the most efficient assert here. This may be more now that you're using this code more. Can be deferred to the future

runtime/libc/uvwasi_backing.c
```diff
@@ -0,0 +1,1403 @@
...
+
+/* Code that actually runs the wasm code */
+IMPORT void wasmf__start(void);
+
+#define unlikely(x) __builtin_expect(!!(x), 0)
+
+static uvwasi_t uvwasi;
+
+/* TODO: This likely makes assumptions about the memory model and needs to be generalized */
+static inline void check_bounds(uint32_t offset, uint32_t bounds_check)
```
Others, line 17: I think you moved this already but didn't remove it here. Also it needs to go in all memory impls

runtime/libc/uvwasi_backing.c
```diff
@@ -0,0 +1,1403 @@
...
+{
+    // Due to how we setup memory for x86, the virtual memory mechanism will catch the error, if bounds < WASM_PAGE_SIZE
+    assert(bounds_check < WASM_PAGE_SIZE || (memory_size > bounds_check && offset <= memory_size - bounds_check));
+}
+
+/**
+ * @brief Lifecycle function that runs at end at termination of WebAssembly module. Suitable for logging and cleanup
+ */
+void runtime_on_module_exit()
+{
```
Others, line 27: The formatting in this file doesn't match the rest of the runtime. I think there is an auto format file in the repo

Others (commented):
Quick comments. This looks fantastic! Will do a full review Sunday

example_code/args/args.c
bushidocodes: In our case, because the WASI exit syscall implementation just calls exit, the main difference is whether the runtime's main function gets returned to or not:
```c
int main(int argc, char* argv[]) {
    runtime_init();
    runtime_args_init(argc, argv);

    atexit(runtime_on_module_exit);

    switch_out_of_runtime();
    wasmf__start();
    /* WASI wrappes non zero status codes in exits, but silently returns on status code 0 */
    return 0;
}
```
A return would return from `wasmf__start` to the main function, but an exit skips this and runs the atexit handlers. Since we have our cleanup logic in the atexit handlers, it basically does the same thing if the modules returns or calls exit.

runtime/memory/64bit_nix.c
bushidocodes: Opened as #49

runtime/libc/uvwasi_backing.c
bushidocodes: I _think_ that may only be in PR #31. I rebased that and marked for review. Happy to rebase and reformat that here once that's approved and merged.

gparmer:
Make sure that you properly attribute the test files to node (and maintain the license). This all looks amazing to me.

gparmer:
It sounds like all subsequent steps are going to build on this. There are a few issues (that you pointed out), and I'm ok with creating issues for those, and progressing with this.

We should all chat about how we want to structure the different Wasi implementations WRT awesome. Seems like we don't want N implementations in this repo (and I think that we all agree with this). Certainly a TODO for the future after we have a chat about it.

PR #52, opened 2021-08-27
feat: WASI Support, Abstract Interface, uvwasi and minimal backings

### LINKED ISSUE #36: Run WASI entrypoint _start instead of main
Currently, the awsm runtime implicitly treats a function called `main` as the entrypoint. 

However, this differs from the WASI standard, which expects this to be `_start`.

I propose that we change this to align to WASI and make it simpler for us to run WASI module.

### LINKED ISSUE #53: Harden or Replace use of VLAs in abstract WASI serialization layer
Currently, the abstract WASI layer uses dynamic memory allocation via VLAs. This has negative security implications for stack overflows and likely will inhibit refactoring of the cortex-m backing to use WASI. This should be addressed and improved to resolve these issues.

bushidocodes:
Resolves #36 
Resolves #53 

1. Provides a WASI abstract interface that allows us to create alternate pluggable backends
2. Implements two different backends, a full-featured one that uses uvwasi and a minimal one written from scratch that only implements a few syscalls
3. Adds WASI unit tests

Related to SLEdge PR at https://github.com/gwsystems/sledge-serverless-framework/pull/267


runtime/libc/wasi/include/wasi_backing.h
```diff
@@ -0,0 +1,1320 @@
...
+ * This implementation ignores client calls and silently returns RC 0
+ *
+ * @return WASI_ESUCCESS
+ */
+uint32_t wasi_snapshot_preview1_sched_yield(void) {
+#ifdef LOG_WASI
+    fprintf(stderr, "wasi_snapshot_preview1_sched_yield()\n");
+#endif
+
+    return (uint32_t)__wasi_sched_yield(CURRENT_WASI_CONTEXT);
```
gparmer, line 1244: Is all of awsm using 4 space indentation?  I thought it was our normal 8. If it all was unified around 4, so be it.

runtime/libc/wasi/include/wasi_serdes.h
```diff
@@ -0,0 +1,364 @@
...
+wasi_serdes_readv_ciovec_t(const void* ptr, size_t end, size_t offset, __wasi_ciovec_t* iovs, __wasi_size_t iovs_len);
+
+static inline __wasi_errno_t
+wasi_serdes_readv_iovec_t(const void* ptr, size_t end, size_t offset, __wasi_iovec_t* iovs, __wasi_size_t iovs_len);
+
+/* Helper functions for memory bounds checking. */
+static inline int wasi_serdes_check_bounds(size_t offset, size_t end, size_t size);
+static inline int wasi_serdes_check_array_bounds(size_t offset, size_t end, size_t size, size_t count);
+
+static inline int wasi_serdes_check_bounds(size_t offset, size_t end, size_t size) {
```
gparmer, line 133: Did the formatter put all of this on the same line? Does awsm not use the "type on line alone" style? Sorry I completely forget.

gparmer: I think that I'm coming to understand that you likely got a lot of this logic from elsewhere? If so, make sure to acknowledge that in the header (and maintain license).

runtime/libc/wasi/wasi_impl_minimal.c
```diff
@@ -0,0 +1,1129 @@
...
+#include <sys/uio.h>
+
+#include "wasi_impl.h"
+
+/*
+ * Essentially based off the uvwasi_t
+ * TODO: Add file descriptors in support of preopens
+ */
+typedef struct wasi_context_s {
+    __wasi_size_t argc;
```
gparmer, line 27: I'd expect some of the previous comments to be blown off since they were to imported code from node. So I'll repeat here: validate that is the 4 space indents are typical in awsm.

runtime/libc/wasi/wasi_impl_minimal.c
```diff
@@ -0,0 +1,1129 @@
...
+    /* Write env and env_buf */
+    __wasi_size_t env_buf_written = 0;
+    for (int i = 0; i < envc; i++) {
+        wasi_context->env[i] = &(wasi_context->env_buf[env_buf_written]);
+        strcpy(wasi_context->env[i], options->envp[i]);
+        env_buf_written += (strlen(options->envp[i]) + 1);
+    }
+
+    /* See Random */
+    srandom(time(NULL));
```
gparmer, line 130: I wouldn't expect these to be available on microcontrollers. Just a comment.

runtime/libc/wasi/wasi_impl_minimal.c
```diff
@@ -0,0 +1,1129 @@
...
+ * @return __WASI_ERRNO_SUCCESS, WASI_EAGAIN, WASI_EWOULDBLOCK, WASI_EBADF, WASI_EFAULT, WASI_EINTR, WASI_EIO,
+ * WASI_EISDIR, or others
+ */
+__wasi_errno_t __wasi_fd_read(void* wasi_context, __wasi_fd_t fd, const __wasi_iovec_t* iovs, size_t iovs_len,
+                              __wasi_size_t* nwritten_retptr) {
+    /* TODO: Should this use readv instead of read? */
+    __wasi_size_t sum = 0;
+
+    for (int i = 0; i < iovs_len; i++) {
+        ssize_t res = read(fd, iovs[i].buf, iovs[i].buf_len);
```
gparmer, line 628: This file isn't a minimal example if it relies on POSIX. Implies that we'll have to implement a separate microcontroller version, which is OK.

gparmer: Make sure that Maorui knows this.

runtime/libc/wasi/wasi_impl_minimal.c
```diff
@@ -0,0 +1,1129 @@
...
+ * Shut down socket send and receive channels.
+ *
+ * @param fd
+ * @param how Which channels on the socket to shut down.
+ * @return status code
+ */
+__wasi_errno_t __wasi_sock_shutdown(void* wasi_context, __wasi_fd_t fd, __wasi_sdflags_t how) {
+    /* similar to `shutdown` in POSIX. */
+    return wasi_unsupported_syscall(__func__);
+}
```
gparmer, line 1129: It isn't clear to me why some functions are implemented, and others aren't. If there's a design/assumptions around this, it would be best to document those, perhaps in a comment at the head of the file?

runtime/libc/wasi/wasi_impl_uvwasi.c
```diff
@@ -0,0 +1,315 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <uvwasi.h>
```
gparmer, line 4: I'm just going to assume this is a straightforward port of uvwasi and do a very light scan.

runtime/libc/wasi/wasi_main.c
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @brief Lifecycle function that runs at end at termination of WebAssembly module. Suitable for logging and cleanup
+ */
+void runtime_on_module_exit() {
+    wasi_context_destroy(current_wasi_context);
+    return;
+}
+
+int main(int argc, char* argv[]) {
```
gparmer, line 22: If this is separate from the backing implementations (uvwasi or minimal) to prevent copying it across both implementations, that's OK. But it will not be used on a microcontroller. I'd hope that it is easy to write a new version of this and the wasi implementation so taht Maorui doesn't need to wrestle with the build system.

tests/wasi/README.md
```diff
@@ -0,0 +1,14 @@
+# WASI Tests
+
+Many of these tests are derived work from the Node.js project
```
gparmer, line 3: Amazing you got all of these working.

gparmer (approved):
Looks good to me. The only high-level concerns are 1. adapting to microcontroller, and 2. the use of arbitrary stack extensions.

I'd like you to help @daviddabono understand 1, and us to have a plan to update for 2.

runtime/libc/wasi/include/wasi_backing.h
bushidocodes: It's been four spaces. Here's a link to the repo before I touched it: https://github.com/gwsystems/aWsm/blob/2d654da6d52a95ec140845c7eb02943db67dfbee/runtime/runtime.c. I believe that SLEdge continued this pattern as well. There are stylistic differences in here compared to SLEdge compared to Composite. My priorities with all this have mostly been just focused on getting clang-format hooked up with an auto-format script and a CI check, and as much as possible, I've tried to match the existing de-facto standards in place to minimize resistance to getting the tooling in place.

bushidocodes: It would be relatively simple to tweak the clang-format setup to make this standardized in the future if you wish.

Others (changes requested):
A few small nits. Mostly to do with maintaining mac support :)

runtime/libc/wasi/include/wasi_backing.h
Others: Rust standard is 4 spaces, so I matched that. C code bases are usually 4 or 8. Composite I think is actually tab intended. My preference is to keep this space indented at 4. The code style is intentionally "rusty" and not supposed to be the same as Composite (that takes a lot of ques from Linux)

Others:
Haven't reviewed the .c code thoroughly. I'm going to defer to Gabes review of that to save everyone some time

runtime/libc/wasi/include/wasi_serdes.h
bushidocodes: The aWsm clang-format is configured to place everything on one line to match Gregor's "Rusty" preferences. The SLEdge config uses "type on line alone."

runtime/libc/wasi/wasi_impl_minimal.c
bushidocodes: All of the derived code from Node.js/uvwasi has downstream changes and has been run through clang-format using the local project config.

runtime/libc/wasi/include/wasi_backing.h
```diff
@@ -0,0 +1,1320 @@
...
+    fprintf(stderr, "wasi_snapshot_preview1_fd_pwrite(%u,%u,%u,%lu,%u)\n", fd, iovs_baseptr, iovs_len, offset,
+            nwritten_retptr);
+#endif
+
+    /* Check Bounds */
+    wasi_serdes_check_bounds(iovs_baseptr, CURRENT_MEMORY_SIZE, iovs_len * WASI_SERDES_SIZE_ciovec_t);
+    wasi_serdes_check_bounds(nwritten_retptr, CURRENT_MEMORY_SIZE, WASI_SERDES_SIZE_size_t);
+
+    /* Copy iovs into memory */
+    __wasi_ciovec_t iovs[iovs_len];
```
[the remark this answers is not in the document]
bushidocodes: I'd have to dig back in to be sure, but I believe that the iovecs need to be copied because the serialization layer does "fixups" to replace memory offsets with host addresses. Perhaps some of these intermediate copied can be removed, but it'd require some concentrated effort. I've added #53 as a follow on effort for this. I understand this is critical, but I still think this should be deferred until after baseline WASI support is merged into aWsm and SLEdge.

example_code/environ/Makefile
```diff
@@ -0,0 +1,27 @@
+WASMCC=/opt/wasi-sdk/bin/clang
+WASMLINKFLAGS=-Wl,--allow-undefined,-z,stack-size=32768,--threads=1
```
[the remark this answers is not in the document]
bushidocodes: Good though. Captured in #54

runtime/libc/wasi/include/wasi_backing.h
```diff
@@ -0,0 +1,1320 @@
...
+}
+
+/**
+ * Get the attributes of a file descriptor.
+ *
+ * @param fd
+ * @param fdstat_retptr the offset where the resulting wasi_fdstat structure should be written
+ * @return status code
+ */
+uint32_t wasi_snapshot_preview1_fd_fdstat_get(__wasi_fd_t fd, __wasi_size_t fdstat_retptr) {
```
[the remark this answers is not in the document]
bushidocodes: After adding LTO and compiling with the minimal backing, it seems that that binaries are 30-50kb. The top-level WASI symbols imported by the original *wasm modules are present, and they've successfully LTOed all of the indirection and function calls stuff away.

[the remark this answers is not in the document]
bushidocodes: Here is a zip with the text formats of the various intermediate stages in our compilation pipeline using the minimal WASI backend and with -O3 and LTO turned on. I picked one of the tests that has the required WASI syscalls implemented. It dynamically links to libc and libm and is disassembled into x64, so not apples-to-apples with the microcontroller stuff, but should give an idea of what the resulting native complexity.
[stdin_example.zip](https://github.com/gwsystems/aWsm/files/7132131/stdin_example.zip)

runtime/libc/wasi/wasi_impl_minimal.c
bushidocodes: Added README documentation and started a discussion with him around this on Slack. 🙌

runtime/libc/wasi/wasi_main.c
bushidocodes: That's correct! It's just used in both places to keep things DRY. I added a comment to this affect to the top of the file and added a section to the README discussing this and providing an example of such a difference in SLEdge due its different execution model. The "build system wrestling" to make this work would mainly be in including the correct runtime translation units during the final compilation step.

runtime/libc/wasi/include/wasi_backing.h
```diff
@@ -0,0 +1,1320 @@
+#pragma once
```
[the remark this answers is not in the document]
bushidocodes: What is your justification for this? Is there a specific compiler that you anticipate us having to use that does not support this pragma? Support matrix is at https://en.wikipedia.org/wiki/Pragma_once. Our toolchain is already tied to LLVM, so I don't really see a reason to not take advantages of some of the advantages of that environment.

install_mac.sh
```diff
@@ -16,20 +16,33 @@ brew install cmake
...
-  curl https://sh.rustup.rs -sSf | bash -s -- -y
+	curl https://sh.rustup.rs -sSf | bash -s -- -y
 fi
 source "$HOME/.cargo/env"
 export PATH="$HOME/.cargo/bin:$PATH"
 
 # Build project
 cargo build --release
+
+# Install WASI-SDK 1 2to /opt/wasi-sdk, the path wasi-sdk assumed by default
```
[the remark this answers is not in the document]
bushidocodes: I pushed some additional changes. The Makefiles and Python bench script now detect the installation location based on the WASI_SDK_PATH environment variable as shown in the WASI-SDK readme. The install script for Mac now uses the path format used by brew for llvm (/usr/local/opt/wasi-sdk). However, you should now be able to theoretically setup wasi-sdk however you want now as long as you set WASI_SDK_PATH in your environment. I still would consider this a "global installation" (since the environment variable is global), but I think this seems to be the happy path for this tooling (it's the pattern used in the README and the dpkg that the maintainers public). Installing in this way also allows us to replace relative paths with environment variables, which is helpful for any follow-on efforts to move the native -> *.wasm compilation stuff into a shared submodule.

install_mac.sh
```diff
@@ -3,33 +3,63 @@ git submodule update --init --recursive
...
+cd wasmception || exit
+make
+cd .. || exit
+
+# Install WASI-SDK if WASI_SDK_PATH not already set
+if [[ -n "${WASI_SDK_PATH}" ]] && [[ -x "${WASI_SDK_PATH}/bin/clang" ]];; then
+	echo "wasi-sdk detected"
+else
+	# Install WASI-SDK 12 to /usr/local/opt/wasi-sdk
+	# This uses the same Mac OS idioms as the LLVM brew installation above
```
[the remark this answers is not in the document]
bushidocodes: I added a -l,--local option to do this for Mac.

runtime/memory/64bit_nix.c
```diff
@@ -52,10 +52,13 @@ i32 instruction_memory_grow(i32 count) {
     return prev_size;
 }
 
+INLINE void check_bounds(u32 offset, u32 bounds_check) {
+    // Due to how we setup memory for x86, the virtual memory mechanism will catch the error, if bounds < WASM_PAGE_SIZE
+    awsm_assert(bounds_check < WASM_PAGE_SIZE || (memory_size > bounds_check && offset <= memory_size - bounds_check));
+}
```
gparmer, line 58: For 64 bit, don't we want to avoid bounds checks? Thought that was the point of the 64 bit version where we can allocate 4GiB. This does not block the PR. Just for discussion after you apply.

gparmer:
This is a hell of a firm foundation to keep building on. I certainly approve. @Others can we move this along. It is holding up other work and deadlines are pending ;-)

code_benches/run.py
```diff
@@ -85,14 +85,14 @@ def getmtime_or_zero(x):
...
+WASMCEPTION_BACKING = "{RUNTIME_PATH}/libc/wasmception_backing.c".format(RUNTIME_PATH=RUNTIME_PATH)
 WASMCEPTION_LINKER_FLAGS = "-Wl,--allow-undefined,-z,stack-size={stack_size},--no-threads,--stack-first,--no-entry,--export-all,--export=main,--export=dummy"
 
-WASI_SDK_PATH = ROOT_PATH + "/wasi-sdk"
+WASI_SDK_PATH = os.environ.get("WASI_SDK_PATH", ROOT_PATH + "/wasi-sdk")
 WASI_SDK_CLANG = WASI_SDK_PATH + "/bin/clang"
 WASI_SDK_SYSROOT = WASI_SDK_PATH + "/share/wasi-sysroot"
 WASI_SDK_FLAGS = "--target=wasm32-wasi -mcpu=mvp -O3 -flto"
-WASI_SDK_BACKING = "wasi_sdk_backing.c"
+WASI_SDK_BACKING = "-ldl -pthread -I{RUNTIME_PATH}/libc/wasi/include -I{RUNTIME_PATH}/thirdparty/dist/include {RUNTIME_PATH}/libc/wasi/wasi_main.c {RUNTIME_PATH}/libc/wasi/wasi_backing.c {RUNTIME_PATH}/libc/wasi/wasi_impl_uvwasi.c {RUNTIME_PATH}/thirdparty/dist/lib/libuvwasi_a.a {RUNTIME_PATH}/thirdparty/dist/lib/libuv_a.a".format(RUNTIME_PATH=RUNTIME_PATH)
```
Others, line 95: Nit: break this up over more than one line

runtime/memory/64bit_nix.c
Others: WASM can address the last byte of memory, which means you need an extra guard page (in case you try and do something like `u64* x = U32_MAX; *x`). This guard page approach only works for free for loadsd that are smaller than a page, for obvious reasons

Others (approved):
Approved. Good work. There are some nitty comments outstanding, but lets bias towards unblocking people and merge this. Would be good to follow up later and see if any comments are low hanging fruit

PR #77, opened 2022-02-17
chore: Update wasmparser

bushidocodes:


src/wasm.rs
```diff
@@ -1380,7 +1384,13 @@ impl WasmModule {
         let mut function_indexes: Vec<u32> = Vec::new();
         loop {
             match p.read() {
-                &ParserState::ElementSectionEntryBody(ref v) => function_indexes.extend(v.iter()),
+                &ParserState::ElementSectionEntryBody(ref v) => {
+                    for elem in v.iter() {
+                        if let ElementItem::Func(idx) = elem {
```
Others, line 1389: What are the other options? Should we at least log here?

src/wasm.rs
```diff
@@ -1336,9 +1336,13 @@ impl WasmModule {
...
         match p.read() {
-            &ParserState::BeginActiveElementSectionEntry(table_id) => {
-                ProcessState::TableElementEntry { table_id }
-            }
+            &ParserState::BeginElementSectionEntry { table, ty: _ } => match table {
+                ElemSectionEntryTable::Active(table_id) => {
+                    ProcessState::TableElementEntry { table_id }
+                }
+                ElemSectionEntryTable::Passive => ProcessState::TableElementSection,
+                ElemSectionEntryTable::Declared => ProcessState::TableElementSection,
```
Others, line 1344: What does this mean?

src/wasm.rs
```diff
@@ -285,23 +285,23 @@ pub enum Instruction {
...
-    SetLocal { index: u32 },
-    TeeLocal { index: u32 },
+    LocalGet { index: u32 },
+    LocalSet { index: u32 },
+    LocalTee { index: u32 },
 
-    GetGlobal { index: u32 },
-    SetGlobal { index: u32 },
+    GlobalGet { index: u32 },
+    GlobalSet { index: u32 },
```
Others, line 293: FYI one reason I created this abstraction layer was to avoid coupling names. If you think this is clearer, more power to you

bushidocodes:
@Others - Appreciate the feedback! Since the API changes so much version to version, I'm going to try to get up to latest before trying to grok the differences beyond trying to get all tests to pass. There doesn't seem to be a changelog or release notes for the library, and they have breaking changes every 3rd or 4th release. Not as bad as other languages due to the type system though!😅

src/wasm.rs
bushidocodes: Added with unimplemented panics.

src/wasm.rs
bushidocodes: Since I'm not 100% sure, I've replaced these arms with unimplemented panics. I'll circle back once we're current.

bushidocodes:
@Others - It turned out that `0.59.0` is a complete rewrite, so I think this a good merge point. I haven't added new functionality. The new code paths associated with new WebAssembly features should panic with some sort of log.

Others (approved):
Ship it!

PR #79, opened 2022-02-21
fix: Set cap to max pages by default

bushidocodes:
We've started to use the `max_pages` global in SLEdge to enforce how large a linear memory should be able to grow. If unspecificed, I believe that the default should be "no cap," which we can achieve by defaulting to the largest possible number of WebAssembly pages.

### DIFF
--- a/src/codegen/memory.rs
+++ b/src/codegen/memory.rs
@@ -17,7 +17,8 @@ use crate::codegen::function::compile_function;
 use crate::codegen::type_conversions::wasm_func_type_to_llvm_type;
 use crate::codegen::ModuleCtx;
 
-const MAX_WEBASSEMBLY_PAGES: u32 = 65536;
+// https://webassembly.github.io/spec/core/valid/types.html#memory-types
+const MAX_WEBASSEMBLY_PAGES: u32 = u32::pow(2, 16);
 
 // We add in globals to tell the runtime how much memory to allocate and startup
 // (And what the max amount of allocated memory should be)
--- a/src/codegen/memory.rs
+++ b/src/codegen/memory.rs
@@ -17,6 +17,8 @@ use crate::codegen::function::compile_function;
 use crate::codegen::type_conversions::wasm_func_type_to_llvm_type;
 use crate::codegen::ModuleCtx;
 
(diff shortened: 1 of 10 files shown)


Others:
Doesn't the runtime already check against 0?

bushidocodes:
@Others - a WebAssembly module can actually have no linear memory whatsoever (pure functions or only uses globals). I would expect zero pages to mean that.

Others:
In that case, the module would have no memories, right? Does the spec allow for a memory to have length zero?

Have you checked none of the bundled runtimes try and eager allocate memory? Assuming you have/this doesn't effect any of the existing runtimes, I'm good to approve

src/codegen/memory.rs
```diff
@@ -17,6 +17,8 @@ use crate::codegen::function::compile_function;
 use crate::codegen::type_conversions::wasm_func_type_to_llvm_type;
 use crate::codegen::ModuleCtx;
 
+const MAX_WEBASSEMBLY_PAGES: u32 = 65536;
```
Others, line 20: Nit: Can we compute this value? Like 2^32 / WASM_PAGE_SIZE? That might be slightly more readable

bushidocodes:
Here's what the spec says about the limits type, which is what the memory accepts two of. My changes are in line with 
image
And here's where it talks about validation, which actually discusses the actual cap
image

bushidocodes:
To my eyes, that seems to suggest that memory with min and max of 0 is valid, but presumably a NOOP.

src/codegen/memory.rs
bushidocodes: I changed it to `const MAX_WEBASSEMBLY_PAGES: u32 = u32::pow(2, 16);` to match the way it's expressed in the spec. That work for you?

bushidocodes:
I went back through the runtimes and removes the special cases for `max_pages == 0`. Probably worth checking my changes to the cortex m `expand_memory` functions. I'm guessing they might have been different because of the cost of division or something. Since the denominator is a power of 2, I would assume it would just end up a right shift.

gparmer:
It is great that this removes the checks on `memory_pages == 0`. Don't see a reason not to merge this!

gparmer (approved):

PR #80, opened 2022-02-21
chore: Silence WASI Unimplemented Errors by default

bushidocodes:
We likely will have to run perf testing before all WASI syscalls are implemented, so we should silence these warnings unless told otherwise.


Others:
Please do this the other way round. The default should not be to silently not do syscalls

Others (approved):

PR #83, opened 2022-02-21
feat: globals_len

bushidocodes:
When using `--runtime-globals`, a runtime needs to be able to know the size of vector it should allocate to store the globals.

With runtime globals on:

```wasm
@globals_len = constant i32 6
define void @populate_globals() {
root:
  call void @set_global_i32(i32 0, i32 4)
  call void @set_global_i32(i32 1, i32 8)
  call void @set_global_i32(i32 2, i32 15)
  call void @set_global_i32(i32 3, i32 16)
  call void @set_global_i32(i32 4, i32 23)
  call void @set_global_i32(i32 5, i32 42)
  ret void
}
```

With runtime globals off:

```wasm
@wasmg_internal_0 = constant i32 4
@wasmg_internal_1 = constant i32 8
@wasmg_internal_2 = constant i32 15
@wasmg_internal_3 = constant i32 16
@wasmg_internal_4 = constant i32 23
@wasmg_internal_5 = constant i32 42
@globals_len = constant i32 0
```

### DIFF
--- a/src/codegen/globals.rs
+++ b/src/codegen/globals.rs
@@ -71,6 +71,12 @@ pub fn insert_globals<'a>(
     }
 }
 
+fn write_globals_len<'a>(llvm_ctx: &'a LLVMCtx, llvm_module: &'a LLVMModule, globals_len: u32) {
+    let max_pages_global =
+        llvm_module.add_global_variable("globals_len", globals_len.compile(llvm_ctx));
+    max_pages_global.set_constant(true);
+}
+
 fn insert_native_globals<'a>(
     opt: &Opt,
     llvm_ctx: &'a LLVMCtx,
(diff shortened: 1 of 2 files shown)


src/codegen/globals.rs
```diff
@@ -71,6 +71,12 @@ pub fn insert_globals<'a>(
     }
 }
 
+fn write_globals_len<'a>(llvm_ctx: &'a LLVMCtx, llvm_module: &'a LLVMModule, globals_len: u32) {
```
Others, line 74: Super duper nit: shouldn't this go after the next function?

Others (approved):

src/codegen/globals.rs
bushidocodes: Happy to reorder however you want. Which functions do you want this between?

bushidocodes: I'm going to address the unit tests in here before merging. I might have missed something.

Others: Just above initializer_to_value?

bushidocodes: Done! Going to take a look at the unit test stuff for this PR shortly.

PR #87, opened 2022-03-23
feat: Call runtime function on wasm unreachable instruction

bushidocodes:
Adds a hook that I can use in SLEdge to prevent assertions within a sandbox from crashing the runtime.


bushidocodes:
I haven't! Since it's error handling code that currently crashed out the runtime, I just did functional tests. Since the translation unit would be calling out to an external function defined in C, I suspect I should add noreturn to the runtime code, right?

PR #93, opened 2022-04-04
Start section

bushidocodes:
This PR implements the `(start)` instruction, which registers a function to run during execution. It takes a function index as follows:

```wasm
(start 2)
````

which generates a call as follows

```llvm
define void @awsm_abi__start_fn() {
entry:
  call void @wasmf_internal_2()
  ret void
}
```

When the start instruction isn't called, the compiler generates a NOP function:

```llvm
define void @awsm_abi__start_fn() {
entry:
  ret void
}
```

Here is the test module in `wasm_apps` I'm using, which is not in this diff:
```wasm
(module
	(global $counter (mut i32) (i32.const 0))

	(func (export "increment") (result i32)
		(global.set $counter (i32.add (global.get $counter) (i32.const 1)))
		(global.get $counter)
	)

	(func $init
		(global.set $counter (i32.const 42))
	)

	(start $init)
	
)
```

The is used in this test, which now runs successfully:
```c
#include "../../runtime/runtime.h"

IMPORT int wasmf_increment(void);

int main(int argc, char* argv[]) {
    runtime_init();

    awsm_assert(wasmf_increment() == 43);
    awsm_assert(wasmf_increment() == 44);
}
```

The first invocation of `wasmf_increment()` returns 1 when I comment out the `(start $init)` instruction because the generated `awsm_abi__start_fn` doesn't call `wasmf_internal_2()`

### DIFF
--- a/runtime/runtime.c
+++ b/runtime/runtime.c
@@ -328,5 +328,8 @@ void runtime_init() {
         runtime_heap_base = memory_size;
     }
 
+    // The start component of a module declares the function index of a start function that is automatically invoked
+    // when the module is instantiated, after tables and memories have been initialized.
+    // https://webassembly.github.io/spec/core/syntax/modules.html#start-function
     awsm_abi__start_fn();
 }
--- a/runtime/runtime.c
+++ b/runtime/runtime.c
@@ -304,6 +304,9 @@ void* allocate_n_bytes_ptr(u32 n) {
 // If we are using runtime globals, we need to populate them
 WEAK void populate_globals() {}
 
+// If a function is registered using the (start) instruction, call it
+WEAK void awsm_abi__start_fn() {}
+
 void runtime_init() {
     if (likely(starting_pages > 0)) {
         alloc_linear_memory();
@@ -324,4 +327,6 @@ void runtime_init() {
     if (runtime_heap_base == 0) {
         runtime_heap_base = memory_size;
     }
+
+    awsm_abi__start_fn();
 }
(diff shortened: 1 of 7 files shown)


runtime/runtime.c
```diff
@@ -329,4 +332,9 @@ void runtime_init() {
     if (runtime_heap_base == 0) {
         runtime_heap_base = memory_size;
     }
+
+    // The start component of a module declares the function index of a start function that is automatically invoked
+    // when the module is instantiated, after tables and memories have been initialized.
+    // https://webassembly.github.io/spec/core/syntax/modules.html#start-function
+    awsm_abi__start_fn();
```
Others, line 339: I have forgotten a bit about the initialization logic. Why isn't this just above where we currently call "main"?

bushidocodes: After messing around with wasm2c and seeing how Mao has been using our runtime, I now tend to think of `main` as outside of our runtime library. That sort of means that `runtime_init` is really something like `module_instantiate`. This basically is [the same pattern as the web](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiate). The key thing is that the wasm `(start)` instruction is basically registering a function to run automatically during instantiation. 

So basically I'm treating `runtime_init` as the wasm instantiation function and placing the start invocation there per the spec: https://webassembly.github.io/spec/core/syntax/modules.html#start-function

This keeps our API clean for a consumer. For example, here's an example program and uses a library compiled as an aWsm module. This doesn't initialize WASI at all because this program doesn't use it.

```c
#include <stdbool.h>

#include "../../runtime/runtime.h"

IMPORT bool wasmf_isArmstrongNumber(int);

int main(int argc, char* argv[]) {
    runtime_init();

    awsm_assert(wasmf_isArmstrongNumber(0) == true);
    awsm_assert(wasmf_isArmstrongNumber(5) == true);
    awsm_assert(wasmf_isArmstrongNumber(10) == false);
    awsm_assert(wasmf_isArmstrongNumber(153) == true);
    awsm_assert(wasmf_isArmstrongNumber(100) == false);
    awsm_assert(wasmf_isArmstrongNumber(9474) == true);
    awsm_assert(wasmf_isArmstrongNumber(9475) == false);
    awsm_assert(wasmf_isArmstrongNumber(9926315) == true);
    awsm_assert(wasmf_isArmstrongNumber(9926314) == false);
}
```

PR #103, opened 2022-09-14
feat: add support for .so dylib files

emil916:
add support for .so dylib files
seperate the wasm instructions from runtime code
implement a missing wasi_intruction (prestat)


.github/workflows/main.yaml
```diff
@@ -34,7 +34,8 @@ jobs:
 
     strategy:
       matrix:
-        libc: [wasmception, wasi-sdk]
+        libc: [wasi-sdk]
+#         libc: [wasmception, wasi-sdk]
```
gparmer, line 38: I don't think that @Others is OK with this.

runtime/memory/64bit_nix.c
```diff
@@ -41,22 +41,10 @@ void expand_memory() {
...
-    }
-
-    return prev_size;
-}
-
 INLINE void check_bounds(u32 offset, u32 bounds_check) {
     // Due to how we setup memory for x86, the virtual memory mechanism will catch the error, if bounds < WASM_PAGE_SIZE
-    awsm_assert(bounds_check < WASM_PAGE_SIZE || (memory_size > bounds_check && offset <= memory_size - bounds_check));
+    awsm_assert(bounds_check < WASM_PAGE_SIZE
+                || (memory_size > bounds_check && offset <= memory_size - bounds_check)); ////// why not offset ?
```
gparmer, line 47: This question is without context. Please add it, and elucidate any concerns you have.

runtime/runtime.c
```diff
@@ -289,28 +21,39 @@ __attribute__((noreturn)) void awsm_abi__trap_unreachable() {
...
-u32 allocate_n_bytes(u32 n) {
-    awsm_assert(memory_size > 0);
-    u32 res = runtime_heap_base;
-    runtime_heap_base += n;
-    while (memory_size < runtime_heap_base) {
-        expand_memory();
+// If a function is registered using the (start) instruction, call it
+WEAK void awsm_abi__start_fn() {}
+
+#ifdef DYNAMIC_LINKING_WASM_SO
```
gparmer, line 27: Is the only way to do this with `ifdef`s? I thought we tried to avoid them in awsm as much as possible. Could be misremembering.

runtime/runtime.h
```diff
@@ -169,4 +173,80 @@ void stub_init();
...
+
+typedef void (*awsm_abi__init_globals_fn_t)(void);
+typedef void (*awsm_abi__init_mem_fn_t)(void);
+typedef void (*awsm_abi__init_tbl_fn_t)(void);
+typedef int32_t (*awsm_abi__entrypoint_fn_t)(void);
+typedef uint32_t (*awsm_abi__wasm_memory_starting_pages_fn_t)(void);
+typedef uint32_t (*awsm_abi__wasm_memory_max_pages_fn_t)(void);
+typedef uint32_t (*awsm_abi__wasm_memory_globals_len_fn_t)(void);
+
+struct awsm_abi_symbols {
```
gparmer, line 186: I can't quite fit it in my head why this is the abi, and other things are not. I think I'm just lacking context here.

runtime/runtime.c
gparmer: Given your later code, is it accurate to say that this is the *complete* set of symbols that an environment depends on an awsm module for?

I don't think it captures all of the symbols that a module depends on an environment for, right?

runtime/runtime.h
```diff
@@ -169,4 +173,80 @@ void stub_init();
...
+static inline int awsm_abi_symbols_init(struct awsm_abi_symbols* abi, const char* path) {
+    awsm_assert(abi != NULL);
+
+    int rc = 0;
+
+    abi->handle = dlopen(path, RTLD_LAZY | RTLD_DEEPBIND);
+    if (abi->handle == NULL) {
+        fprintf(stderr, "Failed to open %s with error: %s\n", path, dlerror());
+        goto dl_open_error;
+    };
```
gparmer, line 206: No `;` here

gparmer (commented):
Looks great! only small issues around 1. wasmception, 2. clarifying the abi wording, and 3. if ifdefs are necessary or if we can just use separate compilation.

runtime/memory/64bit_nix.c
emil916: This was a question to myself. Forgot to remove it :/

runtime/runtime.h
emil916: changed

runtime/runtime.c
emil916: Got rid of the ifdef everywhere...

emil916: ended up separating the wasm instructions for all the architectures in the repo  (mpx, cortex-m ...)

emil916:
I did a complete new cleanup on the repo. 
Got rid of all the #ifdef I added before. Instead provided a simpler Makefile flag to compile for .so lib.
Got rid of a lot of redundancy over the code base.
Made sure it is backward compatible with wasmception, too.

Sorry, it is very a lot to review I guess. But the majority is just the wat files that were missing a line and a keyword change for new version of the wabt app.

gparmer:
@Others You OK with this? Just adding a runtime that can run the awsm-generated code in a .so. Useful for dynamic loading of functions, and for testing this scenario (which is what we do in sledge).

gparmer:
@Others This is an increasingly useful feature to test out interesting support for extensible sandboxes. We're testing strange TLS stuff with it. What do you think about it?

gparmer:
@Others: A little bit more details, and some thoughts. When we use an `extern __thread...` on the variable holding the memory base, function indirection array, and globals, it is generating a ton of calls to `get_tls`. Which means it isn't inlining those values well! This is a pretty significant performance hit. One option to handle this is to simply pass an argument to the struct containing all of these (or 3 separate variables) through all of the functions. This is equivalent, in many ways, to devoting a register to holding it. This would be a significant change to awsm, and is terrifying. But it could really help the dynamic scenario of running sandboxes that aren't LTOed with the runtime, or that use TLS. 

Just a little background. Love to hear you thoughts. I know you're busy; I just wanted to write the above while it is in my head.

